Issue #132, opened 2025-03-28
Compile&Run Llama 3.2 1B backward pass (with Lora layers)


vladimirjovanovicTT:

skhatriTT:
@ndrakulicTT what is our status on this? looks like we can close this one?

Issue #129, opened 2025-03-24
Run GPU training with a subset of examples


vladimirjovanovicTT:
Run GPU training with a subset of examples and store training artifacts so we can compare&iterate quickly.

Issue #125, opened 2025-03-24
Move optimizer to device


vladimirjovanovicTT:

pmarkovicTT:
this is tested and optimizer works on device

Issue #124, opened 2025-03-24
Test various LoRA dimensions


vladimirjovanovicTT:
Currently LoRA dim is 4, we should test for other small dims (8, 16...)

pmarkovicTT:
Tested various dim sizes, up to 1024. All worked fine, the only bottleneck is dram memory for larger size, but smaller sizes should be fine for the initial task

PR #166, opened 2025-05-20
Llama lora pure torch fine tuning

pmarkovicTT:
### Ticket
closes: https://github.com/tenstorrent/tt-blacksmith/issues/132 https://github.com/tenstorrent/tt-blacksmith/issues/124 https://github.com/tenstorrent/tt-blacksmith/issues/125 https://github.com/tenstorrent/tt-blacksmith/issues/129 https://github.com/tenstorrent/tt-blacksmith/issues/123

### Problem description
We need a training script for Llama lora fine tuning with pure PyTorch so we can have fair comparison with GPU training.

### What's changed
Standard PyTorch training loop

### Checklist
- [ ] Add script for pure torch llama lora training


blacksmith/experiments/torch/llama/test_llama_fine_tuning_pure_torch.py
```diff
@@ -0,0 +1,119 @@
...
+                    torch_optimizer.step()
+                    torch_optimizer.zero_grad()
+
+                global_step += 1
+                if global_step % log_every_n_steps == 0:
+                    avg_loss = running_loss / log_every_n_steps
+                    run.log({"train/loss": avg_loss, "step": global_step})
+                    running_loss = 0.0
+
+            if config.save_strategy == "epoch":
```
vladimirjovanovicTT, line 78: what if save_strategy isn't 'epoch' - what are other options? We can develop this a bit; I'm guessing the other options is sth similar to log_every_n_steps, like 'save_every_n_steps'.
We don't want to save a checkpoint probably only if we explicitly set it through a specific save_strategy flag.
Let me know if I'm missing something.

blacksmith/experiments/torch/llama/test_llama_fine_tuning_pure_torch.py
```diff
@@ -0,0 +1,119 @@
...
+    config_file_path = os.path.join(os.path.dirname(__file__), "test_llama_fine_tuning_pure_torch.yaml")
+    config = generate_config(TrainingConfig, config_file_path)
+
+    os.makedirs(os.path.join(config.output_dir, "checkpoints"), exist_ok=True)
+
+    model = get_model(config)
+
+    dataset = SSTDataset(config)
+    train_set, eval_set = dataset.load_tokenized_data()
+    train_data_loader = DataLoader(train_set, batch_size=config.batch_size, shuffle=False, drop_last=True)
```
vladimirjovanovicTT, line 115: I think we should have shuffle=True in train_data_loader.

blacksmith/tools/torch_helpers.py
```diff
@@ -0,0 +1,52 @@
...
+        total_params += param_size
+
+        if param.requires_grad:
+            total_grads += param_size
+
+    # Calculate buffer size (non-parameters that require memory)
+    total_buffers = sum(buf.numel() for buf in model.buffers())
+
+    # Size in bytes = (Number of elements) * (Size of each element in bytes)
+    # We assume parameters and gradients are stored in the same type as input dtype
```
vladimirjovanovicTT, line 35: Optimally we would loop through parameters and gather their dtype but I think this is ok for now.

blacksmith/tools/torch_helpers.py
```diff
@@ -0,0 +1,52 @@
...
+
+    print(f"Input dtype: {input_dtype}")
+    print(f"Model size: {total_memory_gb:.2f} GB")
+    print(f"Parameters: {total_params} | Gradients: {total_grads} | Buffers: {total_buffers}")
+
+    return total_memory_gb
+
+
+def log_mem(stage):
+    allocated = torch.cuda.memory_allocated() / 1e9
```
vladimirjovanovicTT, line 50: nit: we are using 1e9 here and 1024^3 above.

vladimirjovanovicTT (approved):
Looks good, several comments.

blacksmith/datasets/torch/llama/sst_dataset.py
```diff
@@ -4,8 +4,9 @@
 from typing import Tuple, Dict, Any
 from datasets import load_dataset
 from transformers import AutoTokenizer
+import torch
```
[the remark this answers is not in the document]
pmarkovicTT: no that I know, we didn't update hooks recently

PR #245, opened 2025-09-08
Uplift tt-xla to 08.09.2025 version

umalesTT:
Uplifted tt-xla to 08.09.2025 version and deprecated previously used init_device() function for PJRT plugin now that we are using wheel.


ndrakulicTT (approved):
`blacksmith/experiments/jax/mnist/single_chip/test_nnx_mnist.py` and also tensor_paralel examples also have init_device. Besides that LGTM!

Issue #233, opened 2025-08-19
☂️ Training workload for Llama 3.2 1B - LoRA through Jax


vladimirjovanovicTT:
Train Llama 3.2 1B using technique LoRA

This training workload is necessary to showcase tt-xla / forge-fe parity

upantelicTT:
PR merged : https://github.com/tenstorrent/tt-blacksmith/pull/246

PR #246, opened 2025-09-08
Training workload for Llama 3.2 1B - LoRA through Jax

upantelicTT:
https://github.com/tenstorrent/tt-blacksmith/issues/233

☂️ Training workload for Llama 3.2 1B - LoRA through Jax


blacksmith/experiments/jax/llama_jax/README.md
umalesTT, line 1: Couple of things for this file and it's parent directory llama_jax:
- We should keep the naming pattern as it is in other experiments and name parent directory only llama. Then, since we plan on training multiple llama versions, this directory llama should have a subdirectory for llama 3.2 1B LoRA; which is your current llama_jax directory. 
- There are some redundant parts of this readme, in particular "Features", "Setup instructions" (this is covered on a higher level in tt-blacksmith), "Device-specific Differences" and "Dependencies". In general, the amount of details should be similar to other readme files in repo.

blacksmith/experiments/jax/llama_jax/README.md
```diff
@@ -0,0 +1,154 @@
...
+## Usage
+
+### CPU Training
+
+Run LoRA training on CPU:
+
+```bash
+cd blacksmith/experiments/jax/llama_jax
+python training_cpu.py
+```
```
umalesTT, line 62: Should be ```python3 blacksmith/experiments/jax/llama_jax/training_cpu.py```

blacksmith/experiments/jax/llama_jax/README.md
```diff
@@ -0,0 +1,154 @@
...
+
+### TT Device Training
+
+Run LoRA training on Tenstorrent device:
+
+```bash
+cd blacksmith/experiments/jax/llama_jax
+python training_tt.py
+```
```
umalesTT, line 71: Same comment as in cpu case

blacksmith/experiments/jax/llama_jax/lorax/transform.py
```diff
@@ -0,0 +1,126 @@
...
+    def __post_init__(self):
+        assert self.a.shape[-2] == self.b.shape[-1]
+        assert self.w.shape[-2] == self.b.shape[-2]
+        assert self.w.shape[-1] == self.a.shape[-1]
+
+    def materialise(self):
+        return (self.w + self.get_scale() * self.b @ self.a).astype(self.w.dtype)
+
+    def materialize(self):
+        # Alias for British spelling compatibility
```
umalesTT, line 37: nit: I'm not sure I understand? Anyway, no need for both materialise and materialize functions to exist, you can just copy the body of materialise into materialize.

blacksmith/experiments/jax/llama_jax/lorax/transform.py
```diff
@@ -0,0 +1,126 @@
...
+    if lhs_batch or rhs_batch:
+        warnings.warn("Lorax does not support batched matmuls")
+        return False
+    if len(lhs_contract) != 1 or len(rhs_contract) != 1:
+        warnings.warn("Lorax only supports matmul")
+        return False
+    return True
+
+
+@quax.register(lax.dot_general_p)
```
umalesTT, line 58: Can you add a comment on what this decorator is doing?

blacksmith/experiments/jax/llama_jax/lorax/helpers.py
umalesTT, line 1: Given the recent uplifts of jax and python versions, this file will probably need some change to adjust. 
But unrelated to this, please add some comments for each function - quax is not a well-known library and users will have a hard time reading this code (for example, it is not clear what is the difference between merge_params and merge_trainable_frozen, nor what LORA_FREEZE and LORA_FULL actually do)

blacksmith/experiments/jax/llama_jax/lorax/transform.py
umalesTT, line 1: Please add a comment or two for this functions too

blacksmith/experiments/jax/llama_jax/lorax/transform.py
```diff
@@ -0,0 +1,126 @@
...
+
+    second_dimension_numbers = ((lhs_contract,), (0,)), dimension_numbers[1]
+
+    lora_product = op(second, lora_product, dimension_numbers=second_dimension_numbers)
+
+    return (orig + lora_product).astype(orig.dtype)
+
+
+@quax.register(lax.dot_general_p)
+def handle_dot_rhs(lhs: jax.Array, lora: LoraWeight, *, dimension_numbers, **kwargs):
```
umalesTT, line 86: What is the difference between handle_dot_lhs and handle_dot_rhs?

blacksmith/experiments/jax/llama_jax/training_cpu.py
```diff
@@ -0,0 +1,295 @@
...
+        ):
+            print(f"Applying LoRA rank {lora_rank} to: {path_str}")
+            return lora_rank
+        else:
+            return LORA_FREEZE
+
+    return decision_fn
+
+
+def create_loss_fn(lora_model: Any) -> Any:
```
umalesTT, line 118: There is no need for create_loss_fn? Isn't loss_fn sufficient?

blacksmith/experiments/jax/llama_jax/training_cpu.py
```diff
@@ -0,0 +1,295 @@
...
+        model_name=model_name,
+        dataset_id=dataset_id,
+        max_length=max_length,
+        learning_rate=learning_rate,
+        batch_size=batch_size,
+        num_epochs=num_epochs,
+    )
+
+    # Setup wandb for experiment tracking
+    wandb_run = setup_wandb(training_config)
```
umalesTT, line 197: There should be a flag in logging config whether or not to use wandb, because user should be able to execute this program even without wandb access.
In any case I think we should regularly print the results/metrics and if the wandb flag is on, then log to wandb as well.

blacksmith/experiments/jax/llama_jax/training_cpu.py
```diff
@@ -0,0 +1,295 @@
...
+    def train_step(
+        trainable_params: Any,
+        frozen_params: Any,
+        opt_state: Any,
+        input_ids_batch: jnp.ndarray,
+        attention_mask_batch: jnp.ndarray,
+        labels_batch: jnp.ndarray,
+    ) -> Tuple[Any, Any, jnp.ndarray]:
+        loss, grads = jax.value_and_grad(loss_fn, argnums=0)(
+            trainable_params, frozen_params, input_ids_batch, attention_mask_batch, labels_batch
```
umalesTT, line 164: Please add a comment here on why we split params into trainable_params and frozen_params and why we calculate grads only with respect to argnums=0.

blacksmith/experiments/jax/llama_jax/training_tt.py
umalesTT, line 1: All comments for cpu also apply to this tt case.

blacksmith/experiments/jax/llama_jax/training_tt.py
```diff
@@ -0,0 +1,301 @@
...
+            num_batches = len(input_id_batches)
+
+            for batch_idx in range(num_batches):
+                input_ids = input_id_batches[batch_idx]
+                attention_mask = attention_mask_batches[batch_idx]
+                labels = label_batches[batch_idx]
+
+                loss, grads = compute_grads_tt(trainable_params, frozen_params, input_ids, attention_mask, labels)
+
+                with jax.default_device(cpu_device):
```
umalesTT, line 234: Please reference relevant issue here in the comment
Same holds for all other cpu fallbacks - they should be followed by relevant issue which forces that cpu fallback

blacksmith/experiments/jax/llama_jax/training_tt.py
```diff
@@ -0,0 +1,301 @@
...
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import warnings
+from typing import Tuple, Dict, Any, Optional
+
+import jax
+
+jax.config.update("jax_platforms", "tt,cpu")
```
umalesTT, line 11: I think there is no need for this config.update?

blacksmith/experiments/jax/llama_jax/training_cpu.py
```diff
@@ -0,0 +1,295 @@
...
+import os
+import warnings
+from typing import Tuple, List, Dict, Any, Optional
+
+os.environ["TORCH_XLA_DISABLE_DEVICE_PLUGINS"] = "1"
+os.environ["JAX_DISABLE_DEVICE_PLUGINS"] = "1"
+os.environ["TT_BACKEND_DISABLE"] = "1"
+os.environ["DISABLE_TT_PLUGIN"] = "1"
+os.environ["CUDA_VISIBLE_DEVICES"] = ""
+os.environ["JAX_PLATFORMS"] = "cpu"
```
umalesTT, line 15: I think many of these env flags are redundant, my guess is that os.environ["JAX_PLATFORMS"] = "cpu" is sufficient for program to use only cpu

umalesTT (changes requested):
Overall a very good work. Left some comments that mainly address the clarity of the code and structure.

Leaving "Request changes" because of the recent versions uplift of jax and python in our compiler.

blacksmith/experiments/jax/llama/llama_3.2_1B_LoRA/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+Run LoRA training on Tenstorrent device:
+
+```bash
+python3 blacksmith/experiments/jax/llama/llama_3.2_1B_LoRA/training_tt.py
+```
+
+## Configuration Options
+
+Both scripts support the same configurable parameters:
```
ndrakulicTT, line 44: Can we add `.yaml` file similar to `blacksmith/experiments/torch/llama/test_llama_fine_tuning_pure_torch.yaml` that is used in:
`blacksmith/experiments/torch/llama/test_llama_fine_tuning_pure_torch.py`

blacksmith/experiments/jax/llama/lorax/helpers.py
```diff
@@ -0,0 +1,245 @@
...
+    def traverse_tree(params_tree, spec_tree, path=[]):
+        if isinstance(params_tree, dict):
+            for key in params_tree:
+                traverse_tree(params_tree[key], spec_tree[key], path + [key])
+        else:
+            split_param(params_tree, spec_tree, path)
+
+    traverse_tree(lora_params, lora_spec)
+
+    print(f"📊 Split completed:")
```
umalesTT, line 201: Please remove emojis from all prints :)

Issue #236, opened 2025-08-19
☂️ Training workload for Llama 3.2 1B - DoRA


vladimirjovanovicTT:
Train Llama 3.2 1B using technique DoRA

This training workload is necessary to showcase flavor of LoRA finetuning

PR #261, opened 2025-10-14
Training workload for Llama 3.2 1B - DoRA

upantelicTT:
### Ticket
Closes https://github.com/tenstorrent/tt-blacksmith/issues/236

### Model training setting

- **Model:** `Llama 3.2 1B`
- **Strategy:** SFT with DoRA layers on a single chip (N150)
- **Framework:** JAX
- **Dataset:** SST2

### What's changed
Implement DoRA fine-tuning workload for Llama 3.2 1B

### DIFF
--- a/blacksmith/experiments/jax/llama_dora/lorax/helpers.py
+++ b/blacksmith/experiments/jax/llama_dora/lorax/helpers.py
@@ -10,13 +10,10 @@ See THIRD_PARTY_NOTICES.md for the full MIT license text.
 import jax
 import jax.numpy as jnp
 from jax.tree_util import tree_map_with_path, DictKey, SequenceKey
-
-import optax
-import quax
+from typing import Any, Tuple, Dict
 
 from .constants import DORA_FREEZE, DORA_FULL
 from .transform import DoraWeight
-from typing import Any, Tuple, Dict
 
 
 def init_dora(param_tree, spec, rng, stddev=0.01, dtype=jnp.float32, alpha=1.0, is_leaf=None) -> Any:
--- a/blacksmith/experiments/jax/llama_dora/lorax/transform.py
+++ b/blacksmith/experiments/jax/llama_dora/lorax/transform.py
@@ -1,21 +1,21 @@
 # SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
 #
 # SPDX-License-Identifier: Apache-2.0
-from dataclasses import dataclass
-from functools import partial
-import warnings
-
 """
 Portions of this file are derived from 'lorax' by davisyoshida (MIT).
 Copyright (c) 2023 davisyoshida
 Source: https://github.com/davisyoshida/lorax
 See THIRD_PARTY_NOTICES.md for the full MIT license text.
 """
+import warnings
 import jax
 import jax.lax as lax
 import jax.numpy as jnp
 import quax
+
 from typing import Any, Callable
+from dataclasses import dataclass
+from functools import partial
 
 
 def dora(f: Callable[..., Any]) -> Callable[..., Any]:
--- a/blacksmith/experiments/jax/llama_dora/test_llama_fine_tuning_jax.py
+++ b/blacksmith/experiments/jax/llama_dora/test_llama_fine_tuning_jax.py
@@ -1,86 +1,27 @@
 # SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
 #
 # SPDX-License-Identifier: Apache-2.0
-
-import warnings
 import os
-from typing import Tuple, Dict, Any, Optional
-
 import jax
 import jax.numpy as jnp
 import optax
 import numpy as np
-from transformers import FlaxAutoModelForCausalLM, AutoTokenizer, AutoConfig
-
-from datasets import load_dataset
+import wandb
 import lorax
+
+from transformers import FlaxAutoModelForCausalLM, AutoTokenizer, AutoConfig
+from typing import Tuple, Any
+from pathlib import Path
 from lorax import DORA_FREEZE
 
 from blacksmith.datasets.torch.llama.sst_dataset import SSTDataset
 from blacksmith.experiments.torch.llama.configs import TrainingConfig
 from blacksmith.tools.cli import generate_config
-import wandb
-
-MODEL_NAME = "Erland/Llama-3.2-1B-JAX"
-DEFAULT_EXPERIMENT_NAME = "Llama-TT-DoRA-Training"
-DEFAULT_RUN_NAME = "llama-3.2-1b-sst2-tt-lorax"
-
-
-WANDB_ENABLED = False
-
-
-def setup_wandb(training_config: TrainingConfig, enable: bool = False, device: str = "tt") -> Optional[Any]:
-    """Optionally setup wandb for experiment tracking; returns run or None.
-
-    device: one of {"tt", "cpu"}
-    """
-    global WANDB_ENABLED
-    WANDB_ENABLED = bool(enable and (wandb is not None))
-    if not WANDB_ENABLED:
-        return None
-    wandb_run = wandb.init(
-        project=DEFAULT_EXPERIMENT_NAME,
-        name=DEFAULT_RUN_NAME,
-        config={
-            "model_name": training_config.model_name,
-            "dataset_id": training_config.dataset_id,
-            "max_length": training_config.max_length,
-            "learning_rate": training_config.learning_rate,
-            "batch_size": training_config.batch_size,
-            "num_epochs": training_config.num_epochs,
-            "dora_rank": training_config.lora_r,
-            "dora_target_modules": training_config.lora_target_modules,
-            "device": device,
-            "framework": "jax_lorax",
-        },
-    )
-    print(f"Started wandb run: {wandb_run.name}")
-    return wandb_run
-
-
-def log_to_wandb(data_dict: Dict[str, Any], step: Optional[int] = None) -> None:
-    """Log data to wandb if enabled; otherwise no-op."""
-    if WANDB_ENABLED and wandb is not None:
-        wandb.log(data_dict, step=step)
-
-
-def _select_preferred_device() -> Tuple[jax.Device, str]:
-    """Prefer TT device if available, otherwise fall back to CPU.
-
-    Returns (device, device_kind_str)
-    """
-    cpu = jax.devices("cpu")[0]
-    try:
-        tt_devs = jax.devices("tt")
-    except Exception:
-        tt_devs = []
-    if tt_devs:
-        return tt_devs[0], "tt"
-    return cpu, "cpu"
 
 
 def create_batches(data: jnp.ndarray, batch_size: int = 8) -> jnp.ndarray:
     """Create training batches from input data."""
+    data = jnp.array(data)  # Ensure input is a JAX array
     num_batches = len(data) // batch_size
     batched_data = data[: num_batches * batch_size].reshape(num_batches, batch_size, -1)
     return batched_data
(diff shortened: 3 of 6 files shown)


blacksmith/experiments/jax/llama_dora/lorax/helpers.py
```diff
@@ -0,0 +1,172 @@
...
+        m = col_norms.reshape(m_shape).astype(param.dtype)
+
+        return DoraWeight(param, a, b, m, alpha=alpha)
+
+    return jax.tree_util.tree_map_with_path(get_param, param_tree, spec, is_leaf=is_leaf)
+
+
+def simple_spec(params, decision_fn=None, tune_vectors=False, is_leaf=None) -> Any:
+    """
+    Create a simple dora spec for a pytree
```
mmilosevicTT:
```suggestion
    Create a simple DoRA spec for a pytree.
```

blacksmith/experiments/jax/llama_dora/lorax/helpers.py
```diff
@@ -0,0 +1,172 @@
...
+        value = decision_fn(path, arr)
+        return value
+
+    return tree_map_with_path(full_fn, params, is_leaf=is_leaf)
+
+
+def split_trainable_frozen(dora_params, dora_spec) -> Tuple[Dict[str, Any], Dict[str, Any]]:
+    """
+    Split DoRA parameters into trainable and frozen pytrees.
+    Trainable: Only DoRA a,b matrices from DoraWeight objects
```
mmilosevicTT: nit: Please add this comment either in doc or in code, duplication seems unnecessary.

blacksmith/experiments/jax/llama_dora/lorax/transform.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+@quax.register(lax.dot_general_p)
+def handle_dot_lhs_dora(dora: DoraWeight, rhs: jax.Array, *, dimension_numbers: Any, **kwargs: Any) -> Any:
+    """
+    Handle DoRA forward pass: y = (m * normalize(W + B @ A)) @ x
+    """
+    if isinstance(rhs, DoraWeight):
+        rhs = rhs.materialise()
+        warnings.warn("Encountered product of two DoraWeights. Materializing the rhs")
```
mmilosevicTT: Log should come before the call of the code. If something breaks, you would not get the log.

blacksmith/experiments/jax/llama_dora/lorax/transform.py
```diff
@@ -0,0 +1,122 @@
...
+    v_scaled = dora.m * v_normed
+
+    out = op(lhs, v_scaled, dimension_numbers=dimension_numbers)
+    return out.astype(dora.w.dtype)
+
+
+@quax.register(lax.transpose_p)
+def eval_dora_transpose(arg: DoraWeight, *, permutation: Any) -> Any:
+    """
+    Define how a `DoraWeight` behaves under transpose.
```
mmilosevicTT:
```suggestion
    Define how `DoraWeight` behaves under transpose.
```

blacksmith/experiments/jax/llama_dora/test_llama_fine_tuning_jax.py
```diff
@@ -0,0 +1,320 @@
...
+
+import jax
+import jax.numpy as jnp
+import optax
+import numpy as np
+from transformers import FlaxAutoModelForCausalLM, AutoTokenizer, AutoConfig
+
+from datasets import load_dataset
+import lorax
+from lorax import DORA_FREEZE
```
mmilosevicTT: nit: It seems like an overkill to import constant like this just to use it once.

blacksmith/experiments/jax/llama_dora/test_llama_fine_tuning_jax.py
```diff
@@ -0,0 +1,320 @@
...
+def main() -> None:
+    """Main training function with configurable parameters."""
+
+    config_file_path = os.path.join(os.path.dirname(__file__), "test_llama_fine_tuning_jax.yaml")
+    training_config = generate_config(TrainingConfig, config_file_path)
+
+    cpu_device = jax.devices("cpu")[0]
+    current_device, device_kind = _select_preferred_device()
+
+    print(f"Loading Llama 3.2-1B model... Using device: {device_kind} -> {current_device}")
```
mmilosevicTT: Are you actually using `device_kind`, it feels like you are using cpu?

blacksmith/experiments/jax/llama_dora/test_llama_fine_tuning_jax.py
```diff
@@ -0,0 +1,320 @@
...
+    # Initializing model parameters on CPU, since jax.random.normal
+    # is currently not supported on device (https://github.com/tenstorrent/tt-xla/issues/1105).
+    with jax.default_device(cpu_device):
+        model = load_model(training_config.model_name)
+
+    model.params = jax.tree_util.tree_map(lambda x: jax.device_put(x, current_device), model.params)
+
+    wandb_run = setup_wandb(training_config, enable=training_config.model_to_wandb, device=device_kind)
+
+    input_id_batches, attention_mask_batches, label_batches = load_data(training_config)
```
mmilosevicTT: If you print loading of the model, you should print loading of the dataset as well.

blacksmith/experiments/jax/llama_dora/test_llama_fine_tuning_jax.py
```diff
@@ -0,0 +1,320 @@
...
+
+    except Exception as e:
+        print(f"Error during training: {e}")
+        log_to_wandb({"error": str(e), "training_failed": True})
+        raise
+
+    finally:
+        if WANDB_ENABLED and wandb is not None:
+            wandb.finish()
+            print("Finished wandb run")
```
mmilosevicTT:
```suggestion
            print("Finished wandb run.")
```

blacksmith/experiments/jax/llama_dora/test_llama_fine_tuning_jax.py
```diff
@@ -0,0 +1,320 @@
...
+from blacksmith.experiments.torch.llama.configs import TrainingConfig
+from blacksmith.tools.cli import generate_config
+import wandb
+
+MODEL_NAME = "Erland/Llama-3.2-1B-JAX"
+DEFAULT_EXPERIMENT_NAME = "Llama-TT-DoRA-Training"
+DEFAULT_RUN_NAME = "llama-3.2-1b-sst2-tt-lorax"
+
+
+WANDB_ENABLED = False
```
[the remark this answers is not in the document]
pmarkovicTT: same for constants above, make sure to add them to the config (check other experiments for reference)

blacksmith/experiments/jax/llama_dora/test_llama_fine_tuning_jax.py
```diff
@@ -0,0 +1,320 @@
...
+
+def load_model(model_name: str) -> FlaxAutoModelForCausalLM:
+    """Load and configure the Llama model for training."""
+    config = AutoConfig.from_pretrained(model_name)
+    config.use_cache = False
+    config.dtype = jnp.float32
+    return FlaxAutoModelForCausalLM.from_pretrained(model_name, config=config, from_pt=False, dtype=jnp.float32)
+
+
+def load_data(training_config: TrainingConfig) -> Tuple[jnp.ndarray, jnp.ndarray, jnp.ndarray]:
```
pmarkovicTT: Should be part of data loader class, make sure to add data loader as specified above

blacksmith/experiments/jax/llama_dora/test_llama_fine_tuning_jax.py
```diff
@@ -0,0 +1,320 @@
...
+
+    input_id_batches, attention_mask_batches, label_batches = load_data(training_config)
+
+    decision_fn = create_dora_decision_fn(training_config.lora_r)
+    dora_spec = lorax.simple_spec(model.params, decision_fn=decision_fn, tune_vectors=False)
+
+    # Initializing model parameters on CPU, since jax.random.normal
+    # is currently not supported on device (https://github.com/tenstorrent/tt-xla/issues/1105).
+    with jax.default_device(cpu_device):
+        dora_params = lorax.init_dora(model.params, dora_spec, jax.random.PRNGKey(42))
```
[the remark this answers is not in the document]
pmarkovicTT: is it a `seed`?

blacksmith/experiments/jax/llama_dora/test_llama_fine_tuning_jax.py
```diff
@@ -0,0 +1,320 @@
...
+        train_labels.append(np.array(item["labels"]))
+
+    train_input_ids = create_batches(jnp.array(train_input_ids), training_config.batch_size)
+    train_attention_mask = create_batches(jnp.array(train_attention_mask), training_config.batch_size)
+    train_labels = create_batches(jnp.array(train_labels), training_config.batch_size)
+
+    return train_input_ids, train_attention_mask, train_labels
+
+
+def create_dora_decision_fn(dora_rank: int) -> Any:
```
pmarkovicTT: It seems to me this should be part of model loading file? Definitely no reason to be inside training script file

blacksmith/experiments/jax/llama_dora/test_llama_fine_tuning_jax.py
```diff
@@ -0,0 +1,320 @@
...
+    finally:
+        if WANDB_ENABLED and wandb is not None:
+            wandb.finish()
+            print("Finished wandb run")
+
+    print("Testing sentiment classification generation....")
+
+
+if __name__ == "__main__":
+    main()
```
pmarkovicTT: Please use main part/guard to init config, wandb and other configurations, then call` train() ` function (renamed `main` function). Check torch llama fine tuning script.

blacksmith/experiments/jax/llama_dora/test_llama_fine_tuning_jax.py
```diff
@@ -0,0 +1,319 @@
...
+    trainable_params, frozen_params = lorax.split_trainable_frozen(dora_params, dora_spec)
+
+    with jax.default_device(cpu_device):
+        optimizer = optax.adamw(learning_rate=training_config.learning_rate, weight_decay=0.01)
+        trainable_params_cpu = jax.tree_util.tree_map(lambda x: jax.device_put(x, cpu_device), trainable_params)
+        opt_state = optimizer.init(trainable_params_cpu)
+
+    dora_model = lorax.dora(model)
+    loss_fn = create_loss_fn(dora_model)
+    compute_grads_tt = create_compute_grads_fn(loss_fn)
```
umalesTT: nit: When you change the names of functions, please remove "_tt" from function name as well

blacksmith/experiments/jax/llama_dora/test_llama_fine_tuning_jax.py
```diff
@@ -0,0 +1,320 @@
...
+import numpy as np
+from transformers import FlaxAutoModelForCausalLM, AutoTokenizer, AutoConfig
+
+from datasets import load_dataset
+import lorax
+from lorax import DORA_FREEZE
+
+from blacksmith.datasets.torch.llama.sst_dataset import SSTDataset
+from blacksmith.experiments.torch.llama.configs import TrainingConfig
+from blacksmith.tools.cli import generate_config
```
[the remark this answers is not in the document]
mmilosevicTT: Well, the link you provided states this:
```
However, explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose:
```
In my view, adding `blacksmith` to each import is not adding any value whatsoever, as all our files are in the same source directory. On top of that, I do not agree that it is easy to interpret and refactor. If you decide to change the name of the top directory, you have to add additional tens of lines of code to your change. Non-sibling modules stay the same when it comes to explicit and non-explicit notation, you are only removing LCA of imports to keep it compact.

Since this is a nit comment, I am fine with doing it both ways, but am still saying that adding repeated instances of a single value in all imports is not a good practice.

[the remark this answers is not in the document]
pglusacTT: Yes, it does state that, however the example below shows:

> from . import sibling
> from .sibling import example

and we are doing

> from ....datasets.torch.llama.sst_dataset import SSTDataset

which it is not really applicable to that statement.

Renaming the top-level package is rare and handled by a project-wide refactor. IDEs can easily and safely rename your top-level package and make sure everything is correct. File moves are common and absolute imports are safer and better in those scenarios. 

We can't call this a bad practice, it is actually really common and encouraged. 
Read [django](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style#imports) and [pandas](https://pandas.pydata.org/pandas-docs/version/1.2.4/development/code_style.html#imports-aim-for-absolute) guidelines, for example.

I agree that it is not a big deal, but we should have a clear separation when is it okay to use relative imports for our coding guidelines. 

We should take this offline and discuss our coding guidelines further, the sooner we have them the better.

blacksmith/experiments/jax/llama_dora/test_llama_fine_tuning_jax.py
```diff
@@ -0,0 +1,320 @@
...
+MODEL_NAME = "Erland/Llama-3.2-1B-JAX"
+DEFAULT_EXPERIMENT_NAME = "Llama-TT-DoRA-Training"
+DEFAULT_RUN_NAME = "llama-3.2-1b-sst2-tt-lorax"
+
+
+WANDB_ENABLED = False
+
+
+def setup_wandb(training_config: TrainingConfig, enable: bool = False, device: str = "tt") -> Optional[Any]:
+    """Optionally setup wandb for experiment tracking; returns run or None.
```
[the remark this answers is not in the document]
umalesTT: Yes, I agree, I inserted Args and Returns into more complex functions.

PR #263, opened 2025-10-15
Add PyTorch Qwen 2.5 0.5B LoRA finetuning

pmarkovicTT:
### Ticket
close #252 

### Problem description
This PR adds new experiment, specifically:
- **Model:** `Qwen 2.5 0.5B`
- **Strategy:** SFT with LoRA layers on a single chip (N150)
- **Framework:** PyTorch
- **Task:** Instruction following for text-to-SQL

Additionally, it introduces template for writing PyTorch based experiments alongside basic reusable ML training components.

### What's changed
- New experiment folder is added with training script and supporting files.
- Following components are added to the `tools` dir: `ReproducibilityManager`, `CheckpointManager`, `TrainingLogger`, `StorageBackend`.
- `template` dir with template script for PyTorch training is added to the `tools` dir.

Screenshot 2025-10-28 at 10 34 46

### Checklist
- [x] Components implementation
- [x] Experiment implementation
- [x] GPU testing
- [x] TT device testing

Issue #232, opened 2025-08-19
☂️ Training workload for Llama 3.2 1B - LoRA through torch


vladimirjovanovicTT:
Train Llama 3.2 1B using technique LoRA

This training workload is necessary to showcase torch-xla / forge-fe parity

PR #272, opened 2025-10-29
[FOLLOW-UP][TT-XLA][PyTorch] Align SST2 Dataset with New Infra and JAX Implementation

mmilosevicTT:
### Ticket
This is a follow-up of https://github.com/tenstorrent/tt-blacksmith/issues/232

### Problem description
Since new infra was added in https://github.com/tenstorrent/tt-blacksmith/pull/263, our `SSTDataset` is not aligned to it. 

### What's changed
This PR aligns `SSTDataset` to newly added infra. Amongst changes there are:
- Switch of `SSTDataset` to new infra
- Addition of `split` field to `BaseDataset`, to store split of the dataset
- Addition of `collate_fn` to `BaseDataset`, to allow for custom collator
- Fix of dataset size logging in Gemma, Qwen and Llama experiments
- Fix of validation if no tokenizer was provided for Gemma and Llama
- Renaming of `llama` directory in `datasets` to `sst`
- Move of `collate_fn_for_causal_lm` to `tools`
- Addition of `shuffle` and `drop_last` to Qwen, Llama and Gemma experiemnts
- Switch `stanfordnlp/sst2` to `glue/sst2` to align with JAX implementation

### Checklist
As this PR is chaning different experiments, we need to confirm all of them are still working
- [x] Testing of Gemma experiment
- [x] Testing of Qwen experiment
- [x] Testing of FFE Llama experiment
- [x] Testing of XLA Llama experiment
- [x] Testing of JAX implementation for LoRA/DoRA for LLama

### DIFF
--- a/blacksmith/experiments/torch/gemma/README.md
+++ b/blacksmith/experiments/torch/gemma/README.md
@@ -17,11 +17,12 @@ python3 blacksmith/experiments/torch/gemma/test_gemma_finetuning.py
 
 ## Data
 
-The Stanford Sentiment Treebank 2 (SST-2) dataset is a widely-used benchmark for binary sentiment classification.
+GLUE, the General Language Understanding Evaluation benchmark (https://gluebenchmark.com/) is a collection of resources for training, evaluating, and analyzing natural language understanding systems.
+The Stanford Sentiment Treebank consists of sentences from movie reviews and human annotations of their sentiment. The task is to predict the sentiment of a given sentence. It uses the two-way (positive/negative) class split, with only sentence-level labels.
 Each example consists of a sentence from movie reviews labeled as either positive or negative sentiment.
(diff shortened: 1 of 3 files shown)


blacksmith/datasets/torch/sst2/sst2_dataset.py
```diff
@@ -54,8 +54,8 @@ def _tokenize_function(self, example):
         return example
 
     def _prepare_dataset(self):
-        print(f"Loading dataset ({self.config.dataset_id})...")
-        raw_dataset = load_dataset(self.config.dataset_id, split=self.split)
+        print(f"Loading dataset ({self.config.dataset_id}/{self.config.dataset_configuration})...")
```
pmarkovicTT: I would remove print since we should use logger and we log when the dataset is loaded in the training script. If we notice later that we need to log in this, or any other step in the dataset, we can add logger as well, now it seems to me it might be overkill.

PR #279, opened 2025-11-06
Fix Training for PyTorch Tests

mmilosevicTT:
### Ticket
N/A

### Problem description
During PJRT plugin update, it was noticed that PyTorch experiments are failing with an OOM after few batches.

### What's changed
This PR is aiming to fix training of PyTorch experiments. During investigation, several things were found and fixed by this PR:
- Fix for W&B logging, where train/loss was logged and commited, which prevented val/loss to be logged with same step
- Fix for OOM, where we split execution graph to pre-optimizer and optimizer part
- Removal of `xm.optimizer_step(optimizer)` and replacement with `optimizer.step() + torch_xla.sync(wait=True)`as this command was intended to be used for multi-chip purpose

This PR introduces significant **16x performance boost** for PyTorch experiment for Llama, where previously we would around 1 batch/min, while now we are getting 1000 batches per hour (16 batches/min). Other experiments train much faster as well.

Even though this PR fixes issues with OOMs, when we enable validation, we are still getting problems with OOMs at consistent batches. This problem needs to be investigated even further.

### Checklist
- [x] Test Llama experiment
- [x] Test Qwen experiment
- [x] Test Gemma experiment

### DIFF
--- a/blacksmith/experiments/torch/gemma/test_gemma_finetuning.py
+++ b/blacksmith/experiments/torch/gemma/test_gemma_finetuning.py
@@ -8,7 +8,6 @@ import random
 import torch
 from torch.utils.data import DataLoader
 import torch_xla
-import torch_xla.core.xla_model as xm
 import torch_xla.runtime as xr
 from tqdm import tqdm
 from transformers import PreTrainedTokenizer
@@ -150,15 +149,15 @@ def train(
                 if config.use_tt:
                     torch_xla.sync(wait=True)
 
+                do_validation = global_step % config.val_steps_freq == 0
+
                 if global_step % config.steps_freq == 0:
                     avg_loss = running_loss / config.steps_freq if global_step > 0 else running_loss
-                    logger.log_metrics(
-                        {"train/loss": avg_loss}, commit=global_step % config.val_steps_freq != 0, step=global_step
-                    )
+                    logger.log_metrics({"train/loss": avg_loss}, commit=not do_validation, step=global_step)
                     running_loss = 0.0
 
                 # Validation phase
-                if global_step % config.val_steps_freq == 0:
+                if do_validation:
                     avg_val_loss = validate(model, eval_dataloader, loss_fn, device, config, logger, tokenizer)
                     model.train()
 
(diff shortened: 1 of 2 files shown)


blacksmith/experiments/torch/gemma/test_gemma_finetuning.py
```diff
@@ -138,17 +142,17 @@ def train(
...
                 if config.use_tt:
-                    xm.optimizer_step(optimizer)
                     torch_xla.sync(wait=True)
-                else:
-                    optimizer.step()
 
                 if global_step % config.steps_freq == 0:
                     avg_loss = running_loss / config.steps_freq if global_step > 0 else running_loss
-                    logger.log_metrics({"train/loss": avg_loss}, step=global_step)
+                    logger.log_metrics({"train/loss": avg_loss}, commit=global_step % config.val_steps_freq != 0, step=global_step)
```
mmilosevicTT: Gemma experiment does not combine validation with logging of train/loss, so we need to use a special case here for commit.

blacksmith/experiments/torch/gemma/test_gemma_finetuning.py
```diff
@@ -138,17 +142,19 @@ def train(
 
                 # Backward pass
                 loss.backward()
+                if config.use_tt:
+                    torch_xla.sync(wait=True)
 
                 # Update parameters
+                optimizer.step()
```
pmarkovicTT: This is fine for now, we will need to update this for multi-chip later. We will see if we will need another template or just adapt existing one.

Btw make changes to the template as well.

pmarkovicTT (approved):
lgtm, left a few comments

pglusacTT (approved):
Great

blacksmith/experiments/torch/qwen/test_qwen_finetuning.py
```diff
@@ -99,6 +99,8 @@ def train(config: TrainingConfig, device: torch.device, logger: TrainingLogger,
 
                 # Backward pass
                 loss.backward()
+                if config.use_tt:
```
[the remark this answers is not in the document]
mmilosevicTT: Removed

Issue #209, opened 2025-08-19
☂️ Training workload for Qwen/Qwen2.5-1.5B


vladimirjovanovicTT:
Train Qwen/Qwen2.5-1.5B using technique LoRA

This training workload is necessary to showcase Generality

PR #280, opened 2025-11-06
PyTorch Qwen 2.5 1.5B LoRA Finetuning

pmarkovicTT:
### Ticket
close: #209 

### Problem description
This PR adds new experiment, specifically:

- Model: Qwen 2.5 1.5B
- Strategy: SFT with LoRA layers on a single chip (N150)
- Framework: PyTorch
- Task: Instruction following for text-to-SQL

### What's changed
- Add a config for running this model
- Update Qwen model training script with necessary changes to fix validation function and adapt to xla performance improvement.

Screenshot 2025-11-06 at 17 41 05

### Checklist
- [x] CPU device experiment run
- [x] GPU device experiment run
- [x] TT device experiment run

### DIFF
--- a/blacksmith/experiments/torch/qwen/test_qwen_finetuning.py
+++ b/blacksmith/experiments/torch/qwen/test_qwen_finetuning.py
@@ -58,6 +58,9 @@ def validate(
 
             # Predictions
             predictions = shift_logits.argmax(dim=-1)
+            if config.use_tt:
+                torch_xla.sync(wait=True)
+
             num_val_batches += 1
 
             if config.print_examples:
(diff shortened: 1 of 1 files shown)


blacksmith/experiments/torch/qwen/test_qwen_finetuning.py
```diff
@@ -36,15 +45,38 @@ def validate(
...
+            # Shift logits for causal LM: predict next token
+            # logits[:, :-1] predicts tokens at positions 1:
+            shift_logits = logits[:, :-1, :].contiguous()
+
+            # Loss
+            loss = loss_fn(shift_logits.view(-1, model.model.config.vocab_size), labels.view(-1))
             total_val_loss += loss.item()
 
+            # Predictions
+            predictions = shift_logits.argmax(dim=-1)
```
mmilosevicTT: nit: We added syncs here to other experiments, do we want to add it here just to keep consistent? We can remove from all if it turns out to be problematic.

blacksmith/experiments/torch/qwen/test_qwen_finetuning.py
mmilosevicTT: Please add `commit=False` as we will not be logging validation otherwise to wandb.

blacksmith/experiments/torch/qwen/test_qwen_finetuning.py
pmarkovicTT: good point, will add it

PR #315, opened 2025-12-08
fix: check if wandb run is enabled

ndrakulicTT:
### Problem description
in CI wandb is disabled so `run` object is None

### What's changed
Check if wandb run exists

### Checklist
- [ ] For the future we can think if how we want to propagate `log_on_wandb`
- [x] The flax test in basic-tests should pass now: https://github.com/tenstorrent/tt-blacksmith/actions/runs/20037472368


pmarkovicTT (approved):
lgtm
This is fine for jax, while for torch it shouldn't be an issue as wandb can be disabled through config.

Issue #310, opened 2025-12-05
tt-blacksmith CI - V1


ndrakulicTT:
- [ ] custom yaml configs for tests from the table in: #308 
- [x] Implement uplift of tt-xla from tt pypi in nightly
- [x] on Pr - add all mnist experiments, should run on n150/n300
- [x] on Nightly should run on n150/n300/llmbox/p150
- [x] move json from specifying python to just pytest markers (to enable multiple runs on same runner)

Tests only need to not crash to pass at this point

PR #324, opened 2025-12-16
Fix torch tp, and add dp and tp mnist torch to uplift and basic tests presets

ndrakulicTT:
### Ticket
Part of #310 

### Problem description
For torch tp model sharding was not properly specified in yaml

### What's changed
- Fix sharding spec in yaml, and update expected config in pydantic
- Add tp/dp torch mnist tests to both basic-tests and uplift tests

### Checklist
- [x] Run manual uplift Action to check if all tests are passing: https://github.com/tenstorrent/tt-blacksmith/actions/runs/20299777861

### DIFF
--- a/env/xla_requirements.txt
+++ b/env/xla_requirements.txt
@@ -1,5 +1,5 @@
 --extra-index-url https://pypi.eng.aws.tenstorrent.com
-pjrt-plugin-tt==0.7.0.dev20251215
+pjrt-plugin-tt==0.7.0.dev20251217
 flax
 six
 torchdata


blacksmith/experiments/torch/mnist/tensor_parallel/test_mnist_training.py
```diff
@@ -67,6 +67,8 @@ def train(
     # Build model
     model = MNISTLinear(config.input_size, config.hidden_size, config.output_size, bias=config.bias)
     model = model.to(device_manager.device)
+    device_manager.shard_model(model)
```
umalesTT: Shouldn't this happen within each training step?

umalesTT (approved):

PR #356, opened 2025-12-29
add custom yamls for CI tests

ndrakulicTT:
### Ticket
Progress on #310 

### Problem description
Right now if we want to test the experiment we would need to run whole experiment.

### What's changed
Add `TestDataLoaderWrapper` to halt the dataloader after fixed number of steps.

### DIFF
--- a/blacksmith/datasets/torch/mnist/mnist_dataset.py
+++ b/blacksmith/datasets/torch/mnist/mnist_dataset.py
@@ -1,6 +1,9 @@
 # SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
 #
 # SPDX-License-Identifier: Apache-2.0
+
+from typing import overload
+
 import torch
 import torchvision
 from torch.nn import functional as F
@@ -56,6 +59,7 @@ class MNISTDataset(BaseDataset):
     def __len__(self):
         return len(self.dataset)
 
+    @overload
     def _get_dataloader(self) -> DataLoader:
         return DataLoader(
             self.dataset, batch_size=self.config.batch_size, shuffle=self.split == "train", drop_last=True
--- a/blacksmith/datasets/torch/banking77/banking77_dataset.py
+++ b/blacksmith/datasets/torch/banking77/banking77_dataset.py
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-from typing import Dict
+from typing import Dict, overload
 
 import torch
 from torch.utils.data import DataLoader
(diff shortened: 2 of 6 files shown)


blacksmith/datasets/torch/mnist/mnist_dataset.py
```diff
@@ -57,6 +57,7 @@ def __len__(self):
         return len(self.dataset)
 
     def get_dataloader(self) -> DataLoader:
-        return DataLoader(
+        dataloader = DataLoader(
             self.dataset, batch_size=self.config.batch_size, shuffle=self.split == "train", drop_last=True
         )
+        return self._maybe_wrap_test_dataloader(dataloader)
```
agobeljicTT: nit: I'd do something like `_wrap_conditionally_test_dataloader` or `_wrap_test_dataloader_if_test_config`
the word "maybe" made me think about what it means

blacksmith/datasets/torch/torch_dataset.py
```diff
@@ -38,6 +68,22 @@ def __getitem__(self, idx: int) -> Dict:
         """Get a single example from the dataset"""
         pass
 
+    def _maybe_wrap_test_dataloader(self, dataloader: DataLoader) -> DataLoader:
```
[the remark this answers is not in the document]
pglusacTT: I think that wrapping is the way to go here (as opposed to inheritance) and for type hinting we could use the Union type here.

blacksmith/datasets/torch/torch_dataset.py
```diff
@@ -38,6 +68,22 @@ def __getitem__(self, idx: int) -> Dict:
         """Get a single example from the dataset"""
         pass
 
+    def _wrap_test_dataloader_if_test_config(self, dataloader: DataLoader) -> Union[DataLoader, TestDataLoaderWrapper]:
```
pglusacTT: Can we rename this to something like `_clip_test_dataloader`? 
Function names should convey intent and avoid explaining the internal logic.

blacksmith/tools/test_config.py
```diff
@@ -0,0 +1,22 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+"""Test configuration for limiting training runs during testing."""
```
pglusacTT: No need for this docstring, the class has one.

.github/workflows/test-matrix-presets/basic-test.json
```diff
@@ -12,17 +12,17 @@
     {
         "runs-on": "n300",
         "frontend": "xla",
-        "script": "blacksmith/experiments/torch/mnist/test_mnist_training.py"
+        "script": "blacksmith/experiments/torch/mnist/test_mnist_training.py --test-config tests/configs/test_mnist_training_fast.yaml"
```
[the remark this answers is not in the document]
ndrakulicTT: That will be the case, we will have pytest function that will start the subprocess of the training script. We need way to then pass the test config to that subprocess. So devs/users will never in most cases never manually pass `--test-config` it is mostly interface for the pytest

pglusacTT (approved):

PR #368, opened 2026-01-08
Unify requirements.txt for XLA and GPU envs

pglusacTT:
### Problem description
XLA and GPU envs aren't unified where they should be.

### Ticket
N/A

### What's changed
- FFE no longer consumes the common requirements.txt
- Move common dependencies of XLA and GPU to requirements.txt
- Update docs


pglusacTT:
The GPU env is becoming pretty XLA specific. Should we have separate GPU envs, just rename the current one to XLA_GPU, or keep it as is (with some comment in the docs)? 

I think leaving it be is okay for now, let me know what you think. @vladimirjovanovicTT @ndrakulicTT @mmilosevicTT

mmilosevicTT (approved):
LGTM

PR #375, opened 2026-01-12
Add `training_type` to All Experiments

mmilosevicTT:
### Ticket
NA

### Problem description
After adding adapter layers method, I forgot to add `training_type` parameter to all other experiments and we end up not being able to `get_model` properly.

### What's changed
Adding training_type to other configs.

### Checklist
- [ ] Run of experiments to check that they are working properly


umalesTT (approved):
lgtm

PR #378, opened 2026-01-14
Coding Guidelines

pglusacTT:
### Ticket
N/A

### What's changed
Add coding guidelines.

### DIFF
--- a/docs/src/coding-guidelines.md
+++ b/docs/src/coding-guidelines.md
@@ -32,6 +32,33 @@ def process_data(data):
     # Continue processing data
 ```
 
+### Magic Numbers
+Avoid using magic numbers directly in the code. Instead, define them as named constants at the top of the module or in a separate configuration file. This improves code readability and makes it easier to update values in the future.
+
+Bad example:
+```python
+def do_compute(x):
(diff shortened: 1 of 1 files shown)


docs/src/coding-guidelines.md
```diff
@@ -0,0 +1,302 @@
...
+
+Example:
+```python
+from blacksmith.module import MyClass
+from blacksmith.utils.helpers import my_function
+```
+
+Make sure to import only what is necessary to keep the namespace clean.
+
+Avoid:
```
mmilosevicTT: This might be a repeated thing from line 284. Also would be nice to add to the example something where we have more than one method from the lib.

docs/src/coding-guidelines.md
```diff
@@ -0,0 +1,302 @@
...
+```python
+value = 42
+message = f"The answer is {value}."
+```
+
+Bad example:
+```python
+message = f"The answer is {calculate_answer()}."  # Avoid complex expressions
+
+message = "The answer is {value * 2}."  # Avoid using expressions inside f-strings
```
vladimirjovanovicTT: f missing

PR #380, opened 2026-01-15
Improve `env/activate` and GPU Requirements

mmilosevicTT:
### Ticket
NA

### Problem description
We have a problem where PJRT is updated every day, but `source env/activate` does not take into account new versions. Also, GPU setup is out-of-date.

### What's changed
Adding several things:
- Support for re-install on `source env/activate`
- Remove setting of PYTHONPATH from `env/activate`
- Remove re-install of `blacksmith` package from `env/activate` as it is present in `setup.py`
- Update of GPU setup

### Checklist
- [x] Test if setup works for XLA and GPU. Checked lib versions and experiment runs.

### DIFF
--- a/env/activate
+++ b/env/activate
@@ -52,6 +52,22 @@ get_installed_version() {
     $TT_BLACKSMITH_ENV_DIR/bin/pip show "$1" 2>/dev/null | grep -oP '^Version: \K.*' || echo ""
 }
 
+# Function to get index URLs from requirements file
+# Usage: get_index_urls <requirements_file>
+get_index_urls() {
+    local index_url=$(grep -oP '^--index-url\s+\K\S+' "$1" 2>/dev/null)
+    local extra_index_urls=$(grep -oP '^--extra-index-url\s+\K\S+' "$1" 2>/dev/null)
+
+    local result=""
+    if [ -n "$index_url" ]; then
+        result="--index-url $index_url"
+    fi
+    for url in $extra_index_urls; do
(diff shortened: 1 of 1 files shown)


env/activate
```diff
@@ -42,16 +42,56 @@ case $EXPERIMENT_TYPE in
...
+# Function to get installed package version
+# Usage: get_installed_version <package_name>
+get_installed_version() {
+    $TT_BLACKSMITH_ENV_DIR/bin/pip show "$1" 2>/dev/null | grep -oP '^Version: \K.*' || echo ""
+}
+
+# Function to check package version and reinstall if mismatch
+# Usage: check_package_version <package_name> <requirements_file>
+# Returns 0 if versions match or no check needed, 1 if reinstall required
+check_package_version() {
```
mmilosevicTT: Added like this if we decide to add more libs that need to be checked

env/activate
```diff
@@ -20,10 +20,8 @@ case $1 in
     *) echo "Unknown parameter passed: $1"; return 1 ;;
 esac
 
-export SYSTEM_DIST_PACKAGES="/usr/local/lib/python3.11/dist-packages"
-export PYTHONPATH="$(pwd):$(pwd)/tests:$SYSTEM_DIST_PACKAGES"
```
mmilosevicTT: This was incorrect. We were not installing torch nor torch_xla, we would be using ones from `/usr/local/lib/python3.11/dist-packages` which is intrinsically wrong for PJRT install and GPU setup.

PR #389, opened 2026-01-19
Fix @overload issue

agobeljicTT:
### Ticket
N/A

### Problem description
The issue is that `typing.overload` can only be used for expanding or shrinking the number of args and it needs to leave at least one defined (not tagged with @overload) function.

When using inheritance/polymorphism, this is not the case, so we need to remove the tag.

### What's changed
Tag removal


mmilosevicTT (approved):
LGTM, since you checked `@override` is not present in python 3.11

PR #391, opened 2026-01-20
Adjust dtype of MNIST models to match the config

agobeljicTT:
### Ticket
N/A

### Problem description
Currently, only the newest CNN MNIST experiment takes into account that the model should be switched to `torch.bfloat16` (or other depending on the .yaml config).

This PR patches other experiments to do a `model.to(dtype)` correctly.

### What's changed
Added `model.to(dtype)` to each of the MNIST experiments.


mmilosevicTT (approved):
LGTM, left one comment for further compaction of the code.

PR #426, opened 2026-01-27
Add FuseChat for Llama 3.2 3B Training

### LINKED ISSUE #422: Add Different Dataset for Llama 3.2 3B Training Workload
### Description

Add new dataset to TT-Blacksmith which should be used for Llama 3.2 3B Training Workload.

### Goal

Successfully run a complete LoRA finetuning workflow for Llama-3.2-3B through tt-blacksmith, validating the model's compatibility with the training pipeline.

mmilosevicTT:
### Issue
Closes https://github.com/tenstorrent/tt-blacksmith/issues/422

### Problem description
Llama 3.2 3B experiment was added with SST2 dataset results, but we are requiring more involved datasets for models to be considered well impelmented.

### What's Changed
This PR introduces a new dataset - FuseChat.

This PR stumbled upon following issues:
- Validation is not working, we end up in a problem with pcc. This will be fixed as part of https://github.com/tenstorrent/tt-mlir/issues/8666

### Results

After running experiments on GPU and TT we have got these results:
image
image

### Checklist
- [x] Add FuseChat dataset
- [x] Check CPU/GPU runs on Llama 3.2 3B
- [x] Check TT runs on Llama 3.2 3B


umalesTT (approved):
LGTM!
Can you update [experiments table](https://github.com/tenstorrent/tt-blacksmith/blob/main/docs/src/experiments.md) to point to FuseChat instead of sst-2 for this case?

Issue #311, opened 2025-12-05
tt-blacksmith CI - V2


ndrakulicTT:
- [x] Check whole loss and accuracy curve against referent run (only 1st epoch)
  - [ ] If rterr > 0.1 -> notify
  - [x] if final loss rter > 0.5 -> fail job
- [ ] Check if all expected artifacts are dumped in correct place (without saving them)
  - [ ] MLIR graphs
  - [ ] Mesh shape

PR #440, opened 2026-02-04
[CI] Adding golden file functionality for loss checking

agobeljicTT:
### Issue
Part of #311

### Problem description
Current state of CI is that we only check if the exit code is 0.

### What's Changed
- Upload golden files for supported tests.
- Add assertions for loss comparisons with relative tolerance.

### Checklist

### DIFF
--- a/tests/test_all.py
+++ b/tests/test_all.py
@@ -7,9 +7,9 @@ from pathlib import Path
 
 import pandas as pd
 import pytest
+from training_test_cases import TRAINING_TEST_CASES
 
 from blacksmith.tools.logging_manager import GOLDEN_LOGS_DIR, TEST_LOGS_DIR
-from training_test_cases import TRAINING_TEST_CASES
 
 
 @pytest.fixture
--- a/.github/workflows/pr-main.yml
+++ b/.github/workflows/pr-main.yml
@@ -15,7 +15,7 @@ jobs:
     secrets: inherit
 
   uplift-test:
-    if: always()  # REVERT BEFORE MERGING
+    if: ${{ github.head_ref == 'uplift' }}
     uses: ./.github/workflows/call-test.yml
     secrets: inherit
     with:


blacksmith/experiments/torch/gemma/configs.py
```diff
@@ -74,3 +76,4 @@ class TrainingConfig(BaseModel):
     # Other settings
     framework: str = Field(default="pytorch")
     use_tt: bool = Field(default=True)
+    test_config: Optional[TestConfig] = Field(default=None)
```
agobeljicTT: i dont think this pollutes the PR too much, so I'll leave it here

blacksmith/tools/logging_manager.py
```diff
@@ -6,6 +6,7 @@
 import sys
 from typing import Any, Dict, Optional
 
+import pandas as pd
```
ndrakulicTT: I think we need to add this into requirements

tests/test_all.py
```diff
@@ -1,13 +1,56 @@
...
 from training_test_cases import TRAINING_TEST_CASES
 
+LOG_DIR = Path("tests/training_logs")
+GOLDEN_DIR = Path("tests/golden_files")
+
+
+def assert_loss_with_tolerance(log_file: str, golden_file: str, tolerance: float):
+    log_df = pd.read_csv(log_file)
+    golden_df = pd.read_csv(golden_file)
+    pd.testing.assert_frame_equal(log_df, golden_df, rtol=tolerance)
```
ndrakulicTT: Would be good to have a way to have information back where things failed and what is max rtol for example. As in future we would probably want that uploaded to superset

tests/test_all.py
```diff
@@ -70,3 +115,29 @@ def test_training_script(
...
+    if not os.path.exists(os.path.join(GOLDEN_DIR, train_log_file)):
+        # Reference run, move the log files to golden_files.
+        # Sleep to ensure the files are touched after the test run.
+        os.rename(os.path.join(LOG_DIR, train_log_file), os.path.join(GOLDEN_DIR, train_log_file))
+        os.rename(os.path.join(LOG_DIR, val_log_file), os.path.join(GOLDEN_DIR, val_log_file))
+    else:
+        # Test run, compare the train and val log files in training_logs with those in golden_files.
+        assert_loss_with_tolerance(
+            os.path.join(LOG_DIR, train_log_file),
+            os.path.join(GOLDEN_DIR, train_log_file),
```
ndrakulicTT: nit: you can use `/` as both objects are Paths

tests/test_all.py
```diff
@@ -1,13 +1,56 @@
...
+        key=lambda f: f.stat().st_mtime,
+        reverse=True,
+    )
+
+    most_recent_files = log_files[:2]
+    train_log_file = None
+    val_log_file = None
+    for f in most_recent_files:
+        name = f.name
+        if "val" in name or "validation" in name:
```
mmilosevicTT: Second check is redundant here.

tests/test_all.py
```diff
@@ -70,3 +115,29 @@ def test_training_script(
...
         pytest.fail(f"Training script timed out after {setup_dict['timeout']} seconds")
+
+    train_log_file, val_log_file = get_log_files()
+
+    if train_log_file is None or val_log_file is None:
+        return  # If a test does not support golden files yet.
+
+    if not os.path.exists(os.path.join(GOLDEN_DIR, train_log_file)):
+        # Reference run, move the log files to golden_files.
+        # Sleep to ensure the files are touched after the test run.
```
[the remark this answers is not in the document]
agobeljicTT: yep deleted it

tests/golden_files/tt-albert-test_train.csv
```diff
@@ -0,0 +1,26 @@
+_step,train/loss
```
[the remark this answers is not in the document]
agobeljicTT: positive, this was done on a gpu, perhaps it's too little to even see anything

tests/test_all.py
agobeljicTT: done

blacksmith/tools/logging_manager.py
```diff
@@ -21,6 +22,17 @@ def __init__(self, config: TrainingConfig):
         if self.config.use_wandb:
             self._setup_wandb()
 
+        self.test_logs_dir = "tests/training_logs"
+
+        if self.config.test_config is not None:
```
pglusacTT: Not a discussion for this PR, but a general one:
The logger class should clearly use inheritance. It is a set of standardized methods that has a bunch of if statements inside depending on the _type_ of the logger. If we used inheritance we could easily swap out our wandb logger or training logger or whatever logger with the test logger. For now leave this be but it clearly requires a refactor.

blacksmith/tools/logging_manager.py
```diff
@@ -21,6 +22,17 @@ def __init__(self, config: TrainingConfig):
...
+            self.val_log = []
+            self.train_log = []
+
+            os.makedirs(self.test_logs_dir, exist_ok=True)
```
pglusacTT: Avoid using os, use pathlib, see [coding guidelines](https://docs.tenstorrent.com/tt-blacksmith/src/coding-guidelines.html#path-manipulation). If the rest of the file uses os then it just add some comment that it should be refactored please.

blacksmith/tools/logging_manager.py
```diff
@@ -90,6 +104,12 @@ def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None, commi
             except Exception as e:
                 self.std_logger.warning(f"Failed to log to W&B: {e}")
 
+        if self.config.test_config is not None:
+            if "train/loss" in metrics:
+                self.train_log.append({"_step": step, "train/loss": metrics["train/loss"]})
+            if "val/loss" in metrics:
+                self.val_log.append({"_step": step, "val/loss": metrics["val/loss"]})
+
     def log_model_info(self, model_info: Dict[str, Any]):
```
ndrakulicTT: where is `log_model_info` called?

tests/test_all.py
```diff
@@ -70,3 +100,36 @@ def test_training_script(
...
+    test_config = get_run_config(Path(setup_dict["test_config"]))
+
+    run_name = run_config["wandb_run_name"] if "wandb_run_name" in run_config else None
+    if run_name is None:
+        return  # If a test does not support golden files yet.
+
+    train_log_file, val_log_file = get_log_files(run_name)
+
+    if not (TEST_LOGS_DIR / train_log_file).exists() or not (TEST_LOGS_DIR / val_log_file).exists():
+        return  # If a test does not support golden files yet.
```
ndrakulicTT: Hmm I would maybe prefer explicit dont_check_acc flag in test config and then hard asserting on missing files, that way it is expected by default that we check acc, and we are intentionally skipping acc checks for some experiments (jax for example)

tests/test_all.py
```diff
@@ -70,3 +100,36 @@ def test_training_script(
...
+
+    if not test_config["use_tt"] and not (GOLDEN_LOGS_DIR / train_log_file).exists():
```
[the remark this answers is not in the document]
agobeljicTT: Do we want all our experiments to run on the CI?

tests/configs/tt-albert_base_v2-banking77-n150.yaml
```diff
@@ -0,0 +1,10 @@
+test_config:
+  max_steps_per_epoch: 15
+
+steps_freq: 5
+val_steps_freq: 5
+num_epochs: 1
+
+use_tt: true
```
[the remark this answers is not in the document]
agobeljicTT: synced offline
we will put an assertion for true and remove this from configs

PR #444, opened 2026-02-06
Enable forwarding stdout/stderr to system ones in test using `--debug-experiment`

ndrakulicTT:
### Problem description
When running tests, in local there is no live output, which makes it hard to debug tests

### What's Changed
Add simple option to forward the stdout/stderr to system ones using
```sh
pytest --debug-experiment ...
```

### Checklist


tests/test_all.py
```diff
@@ -9,10 +9,16 @@
...
+@pytest.fixture
+def debug(request):
+    return request.config.getoption("--debug-experiment", default=False)
+
+
 @pytest.mark.parametrize("setup_dict", TRAINING_TEST_CASES)
 def test_training_script(
     setup_dict: dict,
     request: pytest.FixtureRequest,
+    debug: bool,
```
agobeljicTT: are we sure this wont break pytest? adding new params here can be tricky, idk if we need to modify test cases or mark.parametrize

is `debug` reserved just like `request` is?

agobeljicTT (approved):

PR #460, opened 2026-02-20
Qwen experiment: removal of do_validate config in favor of steps_freq

agobeljicTT:
> [!CAUTION]
> Replaced by #462 

### Issue
N/A

### Problem description
Currently only Qwen experiment uses `do_validate`, while all other experiments use `steps_freq` to determine if a validation cycle should be run. This PR aims to reintroduce a consistent approach by applying the same check with mod.

### What's Changed
Removed `do_validate` from configs.
Replaced the validation check with a common `global_step % steps_freq == 0`.

### Checklist

### DIFF
--- a/env/xla_requirements.txt
+++ b/env/xla_requirements.txt
@@ -1,4 +1,4 @@
 --extra-index-url https://pypi.eng.aws.tenstorrent.com
 --extra-index-url https://download.pytorch.org/whl/cpu
-pjrt-plugin-tt==0.9.0.dev20260220001050
+pjrt-plugin-tt==0.9.0.dev20260221001203
 torchvision==0.24.0+cpu


blacksmith/experiments/torch/qwen/test_qwen_finetuning.py
```diff
@@ -145,14 +145,15 @@ def train(
...
-                global_step += 1
+                do_validation = global_step % config.steps_freq == 0
+
                 if global_step % config.steps_freq == 0:
                     avg_loss = running_loss / config.steps_freq
                     logger.log_metrics({"train/loss": avg_loss}, commit=False, step=global_step)
                     running_loss = 0.0
 
                     # Do validation
-                    if config.do_validation:
```
mmilosevicTT: Remove this if in total, we should do validation always if we enter the global if.

PR #462, opened 2026-02-21
[Torch] Refactoring experiments to schedule logging and validation consistently across experiments

agobeljicTT:
### Issue
N/A

### Problem description
Currently experiments do not follow a single pattern when it comes to how logging and validation are scheduled.

### What's Changed
This PR introduces a consistent approach with `steps_freq` and `val_steps_freq` to provide more configurability.
Each torch experiment has:
- `global_step` increment moved to the top in order to make the config more intuitive.
- separated validation from train logging, thus making it more configurable and independent
- checkpointing is now independent (the check is done regardless, so no use of putting it within the if of the logging)

### Checklist


blacksmith/experiments/torch/albert/test_albert_finetuning.py
```diff
@@ -116,20 +117,23 @@ def train(
                 # Update parameters
                 device_manager.optimizer_step(optimizer)
 
-                global_step += 1
-                if global_step % config.steps_freq == 0:
+                if (global_step % config.steps_freq == 0) or (global_step == len(train_dataloader)):
```
mmilosevicTT: This will not work for two reasons:
1. Second expression would make it possible for us to enter the if with less then steps_freq losses accumulated. This will lead to faulty log of last step.
2. We will not log anything in cases when we are not in lcm of `steps_freq` and `val_steps_freq` due to `commit=False`.

blacksmith/experiments/torch/albert/test_albert_finetuning.py
```diff
@@ -97,6 +97,7 @@ def train(
     try:
         for epoch in range(config.num_epochs):
             for batch in tqdm(train_dataloader):
+                global_step += 1
```
mmilosevicTT: This technically is correct in the sense of logging, but we are bound to enter errors during train loop if we leave it like this. I prefer increasing this counter only when we actually finished the loop we are adding the counter for.

blacksmith/experiments/torch/llama/xla/test_llama_fine_tuning_pure_torch.py
```diff
@@ -179,16 +181,22 @@ def train(
                     device_manager.optimizer_step(optimizer)
 
                     accumulation_step = 0
-                    global_step += 1
```
mmilosevicTT: Moving this above will mess up the real counter with accumulation steps. We simulate larger batch size, so global_step has to increase only when we reach accumulation steps counter.

blacksmith/experiments/torch/mnist/data_parallel/test_mnist_training_dp.yaml
```diff
@@ -33,6 +31,7 @@ wandb_watch_mode: "all"
 wandb_log_freq: 100
 model_to_wandb: false
 steps_freq: 100
+val_steps_freq: 100
```
mmilosevicTT: This is not the value that was previously present.

mmilosevicTT (changes requested):
I support the idea to unify the way we do logging in all experiments, but please address correctness issues this change would introduce, as current structure was made so we do not miss logs in wandb and to have proper data logged, which is the primary thing we need.

blacksmith/experiments/torch/albert/test_albert_finetuning.py
agobeljicTT: 1. do you think it's ok to remove the second trigger
2. dont we have a flush logic or something? anyway, we can use commit=True always?

blacksmith/experiments/torch/gemma/test_gemma_finetuning.py
```diff
@@ -145,28 +146,20 @@ def train(
...
                 device_manager.optimizer_step(optimizer)
 
-                do_validation = global_step % config.val_steps_freq == 0
-
-                if global_step % config.steps_freq == 0:
-                    avg_loss = running_loss / config.steps_freq if global_step > 0 else running_loss
-                    logger.log_metrics({"train/loss": avg_loss}, commit=not do_validation, step=global_step)
+                if (global_step % config.steps_freq == 0) or (global_step == len(train_dataloader)):
+                    avg_loss = running_loss / config.steps_freq
+                    logger.log_metrics({"train/loss": avg_loss}, step=global_step)
```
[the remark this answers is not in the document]
agobeljicTT: let me run it, I found online that you can do multi logging on the same step.

blacksmith/experiments/torch/llama/xla/test_llama_fine_tuning_pure_torch.py
agobeljicTT: but it'll work for the bottom half of the file if I just revert the increment?

blacksmith/experiments/torch/mnist/data_parallel/test_mnist_training_dp.yaml
agobeljicTT: it was? it was tied to the steps_freq, now I only split those into two ifs, keeping the same value

blacksmith/experiments/torch/llama/xla/test_llama_fine_tuning_pure_torch.py
agobeljicTT: This should be done in another PR

```suggestion
def validate(model, val_data_loader, loss_fn, logger, device_manager, config, tokenizer=None):
```

blacksmith/experiments/torch/albert/test_albert_finetuning.py
```diff
@@ -93,10 +93,18 @@ def train(
...
+        valid_loss, metrics = validate(model, eval_dataloader, logger, device_manager, loss_fn)
+        logger.log_metrics(
+            {"val/loss": valid_loss, "val/accuracy": metrics["accuracy"]},
+            commit=True,
+            step=global_step,
+        )
+
         for epoch in range(config.num_epochs):
             for batch in tqdm(train_dataloader):
+                model.train()
```
mmilosevicTT: I would move this before the loops.

PR #514, opened 2026-04-01
Replace LayerNorm with a custom layer in ALBERT to avoid inf loss

agobeljicTT:
### Issue
Workaround for https://github.com/tenstorrent/tt-metal/issues/41127

### Bug Description
There is a regression in `batch_norm_training` which affects LayerNorm.

### What's Changed
This PR implements a custom TTLayerNorm that should bypass the call to the `batch_norm_training` and force manual ops which (when executed on their own) do not have correctness issues.

### Environment
- **Commit/Version:** N/A
- **Hardware:** N300

### Checklist
- [ ] Bug is reproducible before the fix
- [ ] Fix resolves the bug
- [ ] Tests pass and updated if needed
- [ ] No new bugs introduced
- [ ] Documentation updated if needed


blacksmith/experiments/torch/albert/test_albert_finetuning.py
```diff
@@ -18,6 +18,38 @@
 from blacksmith.tools.device_manager import DeviceManager
 from blacksmith.tools.logging_manager import TrainingLogger
 from blacksmith.tools.reproducibility_manager import ReproducibilityManager
+from blacksmith.tools.workaround_utils import TTLayerNorm
+
+
+def replace_layernorm(module):
```
pglusacTT: This should be placed inside workaround utils.

pglusacTT (approved):
Left only one comment, other than that this is great!
Thanks a lot for doing this, amazing job!

blacksmith/experiments/torch/albert/test_albert_finetuning.py
```diff
@@ -77,6 +109,9 @@ def train(
     if config.resume_from_checkpoint:
         checkpoint_manager.load_checkpoint()
 
+    # TODO(agobeljic): https://github.com/tenstorrent/tt-metal/issues/41127
+    replace_layernorm(model)
```
mmilosevicTT, line 82: Do we need to explicitly unfreeze parameters?

blacksmith/tools/workaround_utils.py
```diff
@@ -37,3 +38,40 @@ def transform_labels(labels, ignored_index, vocab_size):
     expected_output = F.one_hot(labels, num_classes=vocab_size)
 
     return expected_output, labels_mask
+
+
+# Custom LayerNorm for TT-Forge. Necessary because of Albert experiment getting inf loss.
+class TTLayerNorm(nn.Module):
+    def __init__(self, normalized_shape, eps=1e-5, elementwise_affine=True, device=None, dtype=None):
+        super().__init__()
+        # Handle cases where normalized_shape is an int
```
mmilosevicTT: nit: Use full sentences
```suggestion
        # Handle cases where `normalized_shape` is an `int`.
```

mmilosevicTT: Ditto for all other comments.

blacksmith/tools/workaround_utils.py
```diff
@@ -37,3 +38,40 @@ def transform_labels(labels, ignored_index, vocab_size):
     expected_output = F.one_hot(labels, num_classes=vocab_size)
 
     return expected_output, labels_mask
+
+
+# Custom LayerNorm for TT-Forge. Necessary because of Albert experiment getting inf loss.
```
mmilosevicTT: I would put link to the issue without this comment here.

mmilosevicTT (approved):
LGTM

blacksmith/experiments/torch/albert/test_albert_finetuning.py
agobeljicTT: we need to freeze them actually

agobeljicTT: good catch

blacksmith/tools/workaround_utils.py
agobeljicTT: the thing is we don't have a proper issue yet, we haven't found a minimal repro, it isn't trivial.
i can remove the comment nevertheless.

pglusacTT (approved):

PR #525, opened 2026-04-08
Rename and Refactor Experiment Files and Make Optimizers Capturable

agobeljicTT:
### Issue
N/A

### Problem description
Experiment files used inconsistent naming conventions — YAML config files were prefixed with test_ and training scripts were named test_*.py, which conflated training artifacts with test infrastructure and made the directory structure harder to navigate.

### What's Changed
- Renamed all experiment YAML config files: removed test_ prefix and adopted a model_dataset.yaml naming scheme (e.g. test_llama_3_2_1b.yaml → llama_3_2_1b_sst2.yaml)                                                                      
- Renamed all training entry-point scripts from test_*.py to train.py for consistency across torch, JAX, and Lightning experiments
- Reorganized single-chip experiments under a single_chip/ subdirectory                                                                                                                                                                     
- Restructured JAX Llama experiments: moved DoRA and LoRA into llama/dora/ and llama/lora/ subdirectories                                                                                                                                   
- Updated all references in READMEs, CI workflow, and tests/training_test_cases.py to reflect the new paths                                                                                                                                 
- Added new torch/mnist_cnn experiment with CNN training support for MNIST

### Checklist


mmilosevicTT (approved):
LGTM, but we still need to consolidate more things. I would keep this for separate PR.

Summary of things:
Some files have `base_finetuning` others have `fine_tuning`, `finetuning`. Can we consolidate to just remove finetuning from names as well. Also, if we are doing this change we might want to also align to llama experiment structure to have lora dir for yamls as well.

blacksmith/experiments/jax/distil_bert/multi_chip/data_parallel/train.py
```diff
@@ -439,7 +439,7 @@ def train(config: ExperimentConfig, sharding_config: ShardingConfig):
 if __name__ == "__main__":
     jax.config.update("jax_use_shardy_partitioner", True)
 
-    default_config = Path(__file__).parents[2] / "test_distil_bert_flax.yaml"
+    default_config = Path(__file__).parents[2] / "distil_bert_flax_sst2.yaml"
```
pglusacTT: Should the naming schema be model + framework + dataset or just model + dataset?

pglusacTT:
JAX experiments don't seem to be renamed to `train.py`, is that on purpose? Why?

blacksmith/experiments/jax/distil_bert/multi_chip/data_parallel/train.py
agobeljicTT: i think it's ok this way, fw is implied by the the top level parent

agobeljicTT:
> JAX experiments don't seem to be renamed to `train.py`, is that on purpose?

@pglusacTT jes, i skipped jax as they have many diffs between torchs and i thought it would only pollute the PR. perhaps we can tackle it in another one?

pglusacTT:
> > JAX experiments don't seem to be renamed to `train.py`, is that on purpose?
> 
> @pglusacTT jes, i skipped jax as they have many diffs between torchs and i thought it would only pollute the PR. perhaps we can tackle it in another one?

I mean there isn't a big diff structurally, it is always a `train_something.py` coupled with `train_something.yaml`. I don't see a reason for leaving it for another PR.

blacksmith/experiments/jax/llama/dora/README.md
mmilosevicTT, line 1: This is not the same path as we use for torch. I would do `jax/dora/llama` and for the regular `llama` dir, rename to `jax/lora/llama`

blacksmith/experiments/jax/nerf/README.md
mmilosevicTT, line 1: Dataset should be Blender, no?

blacksmith/experiments/torch/mnist/train.py
mmilosevicTT, line 1: I do not like that cnn mnist is just in the same folder as regular mnist `train.py`.

mmilosevicTT (commented):
Almost good to me.

I do not like that only gpt_oss and llama models for hf_models are actually including a dir lora, while others are just plainly just assuming lora. For now this is fine, but if we plan to extend training, this is definitely something we should standardise.

blacksmith/experiments/torch/mnist/train.py
agobeljicTT: wanna make it mnist_cnn and be a top level entry?

blacksmith/experiments/jax/llama/dora/README.md
agobeljicTT: you mean jax/llama/dora and jax/llama/lora? this way is more aligned with torch

blacksmith/experiments/jax/nerf/README.md
agobeljicTT: dataset_name: "Tenstorrent/tt-nerf-p150-white"
in the yaml file

agobeljicTT: ok yes, it should be blender but the yaml is misleading as the train.py just ignores the specified dataset...

blacksmith/experiments/torch/mnist/train.py
mmilosevicTT: it might be good to do that, have no strong opinion about it. @pglusacTT what do you think?

PR #530, opened 2026-04-15
Cursor perf benchmark skill

vladimirjovanovicTT:
## Summary
- Adds a Cursor agent skill for running Tracy device profiling on single-chip training workloads and analyzing results with `tt-perf-report`
- Includes step-by-step instructions for script instrumentation, Tracy capture, and performance analysis

## Files
- `.cursor/skills/perf-benchmark-single-chip/SKILL.md` — main skill procedure
- `.cursor/skills/perf-benchmark-single-chip/reference.md` — metric definitions, CLI flags, output file reference


.cursor/skills/perf-benchmark-single-chip/SKILL.md
```diff
@@ -0,0 +1,247 @@
...
+### 1. Clean device state
+
+Before running any workload, ensure no stale processes are holding the TT device. A previous process that was force-killed (kill -9) leaves the device in a bad state, causing "Timeout waiting for Ethernet core service" or "Waiting for lock 'CHIP_IN_USE_*_PCIe'" errors.
+
+```bash
+ps aux | grep -E "python3.*blacksmith|python3.*torch" | grep -v grep
+
+# Kill any stale processes, then reset both chips:
+tt-smi -r 0
+tt-smi -r 1
```
mmilosevicTT: Why separate resets? `tt-smi -r` should reset all devices. Also, isn't this for single-chip?

.cursor/skills/perf-benchmark-single-chip/SKILL.md
```diff
@@ -0,0 +1,247 @@
...
+```
+
+If `tt-smi` is not installed: `pip install tt-smi`.
```
mmilosevicTT: Should we install this after we start the env?

.cursor/skills/perf-benchmark-single-chip/SKILL.md
```diff
@@ -0,0 +1,247 @@
...
+
+Always reset devices between runs (after a crash or kill) to avoid "Timeout waiting for Ethernet core" errors. Wait a few seconds after reset before starting a new process.
+
+### 2. tt-blacksmith xla venv
```
mmilosevicTT: It is a bit confusing, we are calling env, but we use venv in the title. Can we just name it `Virtual Environment` and then explain that we do it to not mess up other projects running on the same machine.

.cursor/skills/perf-benchmark-single-chip/SKILL.md
```diff
@@ -0,0 +1,247 @@
...
+
+**Tracy tools folder:** Tracy needs `capture-release` and `csvexport-release` binaries. Find them in the workspace:
+
+```bash
+find "$(dirname "$(pwd)")" -name "capture-release" -type f 2>/dev/null
+```
+
+Pick the path under `build_Release/tools/profiler/bin` or similar.
+
+### 4. tt-perf-report
```
mmilosevicTT, line 60: I would add reason why we need this install here as well. I would say explanation like we have for tt-smi give context to agents on why some steps are run and when some potential issues arise.

.cursor/skills/perf-benchmark-single-chip/SKILL.md
```diff
@@ -0,0 +1,247 @@
+---
+name: perf-benchmark-single-chip
+description: Run device performance benchmarks for tt-blacksmith single-chip training workloads. Use when the user asks to benchmark, profile, or measure performance of a training workload on Tenstorrent hardware, or mentions tracy, tt-perf-report, or device time analysis.
+---
+
+# Single-Chip Perf Benchmark
```
mmilosevicTT: nit: Please choose between Title Case and Sentence case

.cursor/skills/perf-benchmark-single-chip/SKILL.md
```diff
@@ -0,0 +1,247 @@
...
+```
+
+**Adapting to different loop structures:**
+
+- If the script uses `torch_xla.sync(wait=True)` -- place `t_start` before the forward call, `t_end` right after the sync.
+- If the script has no explicit sync -- add `torch_xla.sync(wait=True)` after backward and before `t_end`.
+
+**Gradient accumulation:** If the loop accumulates gradients over multiple micro-steps before an optimizer step, place the signpost/timer around each individual fwd+bwd+sync micro-step (not the full accumulation cycle). The early stopping check (step 4 below) should go inside the accumulation-complete branch where `global_step` is incremented.
+
+4. **Add a timer summary function and early stopping** after ~5 steps (keeps the Tracy trace small):
```
mmilosevicTT, line 113: We need to make sure the accumulation num of steps i lower than this, so we can capture optimizer as well, no?

.cursor/skills/perf-benchmark-single-chip/SKILL.md
```diff
@@ -0,0 +1,247 @@
...
+5. **Disable `xr.clear_computation_cache()`** -- If the training script calls `xr.clear_computation_cache()` between steps, comment it out. It forces XLA to recompile the graph every step, which invalidates steady-state measurements and can cause OOM.
+
+**Important:**
+- Step 1 includes XLA graph compilation -- exclude it from steady-state averages.
+
+### Step 2: Run Tracy
+
+```bash
+source env/activate --xla
+tt-smi -r 0 && tt-smi -r 1
```
mmilosevicTT: ditto on reset, I think we should only call `tt-smi -r` or at most `tt-smi -lr`

.cursor/skills/perf-benchmark-single-chip/SKILL.md
```diff
@@ -0,0 +1,247 @@
...
+
+## Troubleshooting
+
+- **"Timeout waiting for Ethernet core service" / "Waiting for lock 'CHIP_IN_USE_*_PCIe'"**: Kill all stale Python processes, then reset both chips: `tt-smi -r 0 && tt-smi -r 1`. Wait a few seconds before starting a new process.
+- **Tracy `ImportError: cannot import name 'setup_tt_metal_home' from 'pjrt_plugin_tt'`**: The installed `pjrt_plugin_tt` wheel is out of sync with the source tree. Fix by adding the tt-xla source package to PYTHONPATH and using `python3 -m tracy` instead of bare `tracy`. See [Tracy setup](#3-tracy).
+- **Tracy `No device operations found`**: Default analysis uses ops after the last signpost. Use `--start-signpost` and `--end-signpost` flags to select a specific step range.
+- **`env/activate --xla` triggers a slow package upgrade**: The activate script auto-detects pjrt-plugin-tt version mismatches and runs `pip install`, which can take 1-2 minutes and churn package versions. Avoid re-sourcing the activate script between tracy and tt-perf-report steps if the venv is already active.
```
mmilosevicTT: I do not think this is a good idea to say like this. We should encourage to upgrade env, as we want to use current wheel when we run training, or at least let agent know that upgrading might help with fixing.

.cursor/skills/perf-benchmark-single-chip/reference.md
```diff
@@ -0,0 +1,141 @@
...
+
+| Bound | Meaning | Threshold |
+|---|---|---|
+| **DRAM** | Memory-bandwidth bound (>65% peak BW) | DRAM % > 65 |
+| **FLOP** | Compute bound (>65% peak FLOPs) | FLOPs % > 65 |
+| **BOTH** | Saturating both DRAM and compute | Both > 65% |
+| **SLOW** | Neither DRAM nor compute saturated | Both < 65% -- **biggest optimization target** |
+| **HOST** | Bottlenecked by host dispatch | Op-to-op gap dominates |
+
+## Wormhole peak numbers
```
mmilosevicTT: Do we have Blackhole numbers?

.cursor/skills/perf-benchmark-single-chip/reference.md
```diff
@@ -0,0 +1,141 @@
...
+| HiFi4 | 74 | Full precision |
+| HiFi2 | 148 | Half precision mantissa |
+| LoFi | 262 | Lowest precision |
+
+- **Peak DRAM BW**: 288 GB/s
+- **Max cores**: 64 Tensix cores
+
+## tt-perf-report CLI flags
+
+```
```
mmilosevicTT: I would maybe just add the command you run to get this, like `tt-perf-report --help`?

mmilosevicTT (approved):
Amazing to see skill additions in this repo as well.

LGTM, left some comments that I believe would improve the work of the agent

.cursor/skills/perf-benchmark-single-chip/SKILL.md
vladimirjovanovicTT: great point, updated.

.cursor/skills/perf-benchmark-single-chip/reference.md
```diff
@@ -0,0 +1,154 @@
...
+--csv <path>         Export enriched per-op CSV + stacked aggregate CSV + PNG chart
+--min-percentage N   Hide ops contributing <N% of total device time (default: 0)
+--id-range M-N       Analyze only ops with IDs in [M, N] (useful for single-layer analysis)
+--start-signpost S   Analyze ops after signpost S (Tracy signpost name)
+--end-signpost S     Analyze ops before signpost S (Tracy signpost name)
+```
+
+## Tracy CLI flags
+
+```
```
mmilosevicTT: Can we add command like we added in tt-perf-report?

.cursor/skills/perf-benchmark-single-chip/reference.md
```diff
@@ -0,0 +1,154 @@
...
+
+| Bound | Meaning | Threshold |
+|---|---|---|
+| **DRAM** | Memory-bandwidth bound (>65% peak BW) | DRAM % > 65 |
+| **FLOP** | Compute bound (>65% peak FLOPs) | FLOPs % > 65 |
+| **BOTH** | Saturating both DRAM and compute | Both > 65% |
+| **SLOW** | Neither DRAM nor compute saturated | Both < 65% -- **biggest optimization target** |
+| **HOST** | Bottlenecked by host dispatch | Op-to-op gap dominates |
+
+## Wormhole peak numbers (single chip)
```
mmilosevicTT: nit: In Blackhole we mentioned P150, why not mention N150 here?

.cursor/skills/perf-benchmark-single-chip/SKILL.md
```diff
@@ -0,0 +1,254 @@
...
+
+step_ms = (t_end - t_start) * 1000.0
+step_times.append(step_ms)
+tag = "COMPILE" if len(step_times) == 1 else "steady"
+print(f"[TIMER] Step {len(step_times)}: fwd+bwd = {step_ms:.1f} ms  ({tag})")
+```
+
+**Adapting to different loop structures:**
+
+- If the script uses `torch_xla.sync(wait=True)` -- place `t_start` before the forward call, `t_end` right after the sync.
```
mmilosevicTT: This comment might not be right. Having loss.item() in code forces sync, but if we placed an additional sync just to be sure, after forward, I believe this would put t_end before backward pass.

.cursor/skills/perf-benchmark-single-chip/SKILL.md
```diff
@@ -0,0 +1,254 @@
...
+- If the script has no explicit sync -- add `torch_xla.sync(wait=True)` after backward and before `t_end`.
+
+**Gradient accumulation:** If the loop accumulates gradients over multiple micro-steps before an optimizer step, place the signpost/timer around each individual fwd+bwd+sync micro-step (not the full accumulation cycle). The early stopping check (step 4 below) should go inside the accumulation-complete branch, **immediately after `global_step` is incremented** and before any validation, logging, or checkpointing. Make sure `BENCH_MAX_STEPS` is large enough to include at least one full optimizer step (i.e., `BENCH_MAX_STEPS >= gradient_accumulation_steps + 1`) so the trace captures the optimizer update as well.
```
mmilosevicTT, line 111: I just realised, if we put t_end after backward, we will never actually trace optimizer, so the comment might be redundant?

PR #541, opened 2026-04-22
Add Support for FSDP

pglusacTT:
### Issue
N/A

### Problem description
Per-device DRAM is becoming a bottleneck for large-scale experiments.
We need to reduce DRAM usage in experiments using Torch-XLA FSDP sharding.

### What's Changed
- Add FSDP wrapper to device manager
- Add logic for resolving strategy

### Swiss report
Megatron:
image

FSDP:
image

## Note: 
There is an issue with loss values we're observing that seems unrelated to this change.
When we aren't getting infs, the loss curve looks fine and loss values are in range.
Adding this PR to enable other experiments to utilize FSDP.
Tracking issue: https://github.com/tenstorrent/tt-blacksmith/issues/542

W&B report:
image

### Checklist

### DIFF
--- a/blacksmith/experiments/torch/gpt_oss/lora/loudbox/fsdp_gpt_oss_20b_sst2.yaml
+++ b/blacksmith/experiments/torch/gpt_oss/lora/loudbox/fsdp_gpt_oss_20b_sst2.yaml
@@ -1,0 +1,64 @@
+# Dataset settings
+dataset_id: "sst2"
+
+# Model settings
+model_name: "openai/gpt-oss-20b"
+max_length: 128
+dtype: "torch.bfloat16"
+
+# Training hyperparameters
+training_type: "lora"
+learning_rate: 1e-4
+gradient_accumulation_steps: 4
+batch_size: 1
+num_epochs: 1
+ignored_index: -100
+
+# LoRA setup
+lora_r: 32
+lora_alpha: 64
+lora_target_modules: ["q_proj", "k_proj", "v_proj", "o_proj"]
+lora_task_type: "CAUSAL_LM"
+
+# Device settings
+mesh_shape: [1, 8]
+mesh_axis_names: ["batch", "fsdp"]
+input_sharding_dim: null
+
+# Reproducibility settings
+seed: 23
+deterministic: False
+
+# Logging settings
+log_level: "INFO"
+use_wandb: True
+wandb_project: "gpt-oss-20b-lora"
+wandb_run_name: "tt-gpt-oss-20b-test"
+wandb_tags: ["test"]
+wandb_watch_mode: "all"
+wandb_log_freq: 1000
+model_to_wandb: False
+steps_freq: 1
+val_steps_freq: 10
+epoch_freq: 1
+
+# Checkpoint settings
+resume_from_checkpoint: False
+resume_option: "last"
+checkpoint_path: ""
+checkpoint_metric: "eval/loss"
+checkpoint_metric_mode: "min"
+keep_last_n: 2
+keep_best_n: 1
+save_strategy: "step"
+project_dir: "blacksmith/experiments/torch/gpt_oss"
+save_optim: True
+storage_backend: "local"
+sync_to_storage: False
+load_from_storage: False
+remote_path: ""
+
+# Other settings
+framework: "pytorch"
+print_examples: True
+use_tt: True
--- a/blacksmith/experiments/torch/gpt_oss/lora/loudbox/fsdp_gpt_oss_20b_sst2.yaml
+++ b/blacksmith/experiments/torch/gpt_oss/lora/loudbox/fsdp_gpt_oss_20b_sst2.yaml
@@ -1,3 +1,6 @@
+# DISCLAIMER: This produces -inf in the loss.
+# TODO(pglusac): https://github.com/tenstorrent/tt-blacksmith/issues/542
+
 # Dataset settings
 dataset_id: "sst2"
 
--- a/blacksmith/experiments/torch/gpt_oss/train.py
+++ b/blacksmith/experiments/torch/gpt_oss/train.py
@@ -196,8 +196,6 @@ def train(
                 running_loss += loss_.item()
                 accumulation_step += 1
 
-                logger.info(f"Current loss and step: {loss_.item()} {global_step}")
-
                 # Only step the optimizer after accumulating gradients.
                 if accumulation_step == config.gradient_accumulation_steps:
                     device_manager.optimizer_step(optimizer)
--- a/blacksmith/experiments/torch/gpt_oss/train.py
+++ b/blacksmith/experiments/torch/gpt_oss/train.py
@@ -112,7 +112,7 @@ def train(
     logger.info("Starting training...")
 
     # Load model.
-    model = get_model(config, device_manager.device, device_manager)
+    model = get_model(config, device_manager, shard_model=True)
 
     logger.info(f"Loaded {config.model_name} model.")
     logger.info(f"Model parameters: {sum(p.numel() for p in model.parameters())}")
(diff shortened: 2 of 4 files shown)


mmilosevicTT (approved):
LGTM

blacksmith/tools/device_manager.py
```diff
@@ -125,6 +134,22 @@ def _apply_tensor_parallelism(self, model: nn.Module) -> nn.Module:
         torch_xla.sync(wait=True)
         return model
 
+    def _apply_fsdp(self, model: nn.Module) -> nn.Module:
+        # TODO(pglusac): Add support for FSDP granularity configuration.
+        def shard_output(output, mesh):
+            # Extract logits and shard its batch dimension along the fsdp axis.
+            real_output = getattr(output, "logits", None)
+            if real_output is None:
+                real_output = output[0] if isinstance(output, tuple) else output
```
mmilosevicTT, line 143: Does it have to be a tuple? Can it ba a list?

blacksmith/tools/device_manager.py
```diff
@@ -94,13 +97,19 @@ def is_tensor_parallel(self) -> bool:
...
+        return self.mesh is not None and "fsdp" in self.mesh.axis_names
+
     def shard_tensor(self, tensor: torch.Tensor, sharding_spec: Tuple):
         return xs.mark_sharding(tensor, self.mesh, sharding_spec)
 
     def shard_model(self, model: nn.Module) -> nn.Module:
         """Shard model based on mesh configuration."""
+        if self.is_fsdp():
+            model = self._apply_fsdp(model)
         if self.is_tensor_parallel():
```
mmilosevicTT, line 111: Are we sure fsdp will shard each layer? Should we do elif just in case?

mmilosevicTT (commented):
Left two questions just to confirm.

blacksmith/experiments/torch/gpt_oss/lora/loudbox/fsdp_gpt_oss_20b_sst2.yaml
```diff
@@ -0,0 +1,67 @@
+# DISCLAIMER: This produces -inf in the loss.
```
agobeljicTT, line 1: is there a case where it wont inf?

blacksmith/models/torch/gpt_oss/model_overrides.py
```diff
@@ -9,8 +9,12 @@
 from transformers.utils.quantization_config import Mxfp4Config
 
 
-def get_model(config, device):
-    """Load GPT-OSS model with deinterleaving overrides, LoRA, and compilation."""
+def get_model(config, device, device_manager=None):
```
agobeljicTT: can we just use device_manager.device? device be redundant here

blacksmith/models/torch/gpt_oss/model_overrides.py
```diff
@@ -40,6 +44,9 @@ def get_model(config, device):
 
     model.to(device)
 
+    if device_manager is not None:
+        model = device_manager.shard_model(model)
```
agobeljicTT, line 49: do we need to remove shard_model from other experiments?

blacksmith/experiments/torch/gpt_oss/train.py
```diff
@@ -180,7 +178,6 @@ def train(
                 }
                 # Shard batch if data parallelism is used.
                 batch = device_manager.prepare_batch(batch)
-                device_manager.shard_model(model)
```
agobeljicTT, line 183: so this was done on each cycle of training?

blacksmith/tools/device_manager.py
pglusacTT: I deliberately put it this way. FSDP will do it's thing, shard the entire model but you still have a choice to override shardings.

blacksmith/tools/device_manager.py
pglusacTT: Sure, will extend.

blacksmith/experiments/torch/gpt_oss/lora/loudbox/fsdp_gpt_oss_20b_sst2.yaml
pglusacTT: I put up the w&b logs. You can see that it infs for some steps, not for others.

blacksmith/models/torch/gpt_oss/model_overrides.py
pglusacTT: I want this to be optional for now, might do it later.

blacksmith/experiments/torch/gpt_oss/train.py
pglusacTT: Yes but it doesn't seem to be necessary.

blacksmith/models/torch/gpt_oss/model_overrides.py
agobeljicTT: add a TODO or a comment so we know this is deliberate

blacksmith/experiments/torch/gpt_oss/train.py
umalesTT, line 139: Should we put this in yaml file?

umalesTT: nit: we don't have per-step logging in other experiments, i guess this got left from debugging

blacksmith/models/torch/huggingface/hf_models.py
```diff
@@ -27,6 +27,9 @@ def get_model(config: TrainingConfig, device: torch.device):
     model.to(eval(config.dtype))
     model.to(device)
 
+    if device_manager is not None:
```
umalesTT: Can we do `get_model(config, device_manager, some_flag)`?
If I'm not mistaken you use `device_manager` as that flag here to make this optional - if we come up with another name we won't have redundant `device` arg and a more clear name for the flag

blacksmith/tools/device_manager.py
umalesTT: For my understanding, if we wanted every odd layer with FSDP, we would FSDP the whole model and override shardings of even layers?

blacksmith/tools/device_manager.py
```diff
@@ -125,6 +134,22 @@ def _apply_tensor_parallelism(self, model: nn.Module) -> nn.Module:
...
+            real_output = getattr(output, "logits", None)
+            if real_output is None:
+                real_output = output[0] if isinstance(output, tuple) else output
+            # Skip if the sharding annotation is already set.
+            if torch_xla._XLAC._get_xla_sharding_spec(real_output) not in (None, ""):
+                return
+            partition_spec = ("fsdp",) + (None,) * (real_output.dim() - 1)
+            xs.mark_sharding(real_output, mesh, partition_spec)
+
+        model = FSDP(model, mesh=self.mesh, shard_output=shard_output)
```
umalesTT, line 151: Why do we shard output by default?
I feel like we should have an option to do FSDP only , irregardless of output sharding

umalesTT (approved):
🔥

blacksmith/experiments/torch/gpt_oss/train.py
pglusacTT: I don't think it should reside in the yaml file. This is something that is model specific and not related to the experiment itself.

blacksmith/models/torch/huggingface/hf_models.py
pglusacTT: Makes sense, will do it.

blacksmith/tools/device_manager.py
pglusacTT: Yes, we could do that as well, didn't think of it!

blacksmith/tools/device_manager.py
pglusacTT: Ahhh it is buggy when sharding output is not provided. I'll add a todo for it but yeah.

PR #558, opened 2026-05-06
Add run_decode_example which runs real decode loop during validation instead of showing teacher forcing examples

mcupacTT:
### Problem Description
Validation loop did not have functionality to run a full decode apart from teacher forcing decode (only predicting next token from the ground truth).

### PR description
This PR adds `run_decode_example` running a decode loop and using static KV cache. The code is adopted from [tt-xla](https://github.com/tenstorrent/tt-xla/blob/main/examples/pytorch/llama.py)

`run_decode_example` is invoked from `run_decode_example_from_batch`, which is extracting prompt and target token ids from the combined items which are passed from the data loader.

Currently, the functionality is only implemented within gemma11.

### Example, tests, and timing
This is how it looks like in the code:
image
Since we don't have `EOS` in sst2 template, it didn't stop generating after finishing sentiment info, which is expected.

On first validation, the decode takes ~50sec, but for every following validation it is 5sec or less. 

### Side note
I didn't put any other `torch.compile` in the decode function apart from the initial one for the training (this can be discussed)

### DIFF
--- a/blacksmith/experiments/torch/gemma11/train.py
+++ b/blacksmith/experiments/torch/gemma11/train.py
@@ -20,8 +20,8 @@ from blacksmith.tools.reproducibility_manager import ReproducibilityManager
 from blacksmith.tools.torch_helpers import (
     collate_fn_for_causal_lm,
     collect_examples,
-    show_examples,
     run_decode_example_from_batch,
+    show_examples,
 )
 
 
@@ -70,7 +70,7 @@ def validate(model, val_data_loader, loss_fn, device_manager, config, logger, to
     if config.print_examples and tokenizer is not None:
         logger.info(f"\n=== Validation Examples (Random samples) ===")
         show_examples(collected_examples, tokenizer, config, logger)
-    
+
     if config.run_decode_example and tokenizer is not None:
         run_decode_example_from_batch(
             model=model,
--- a/blacksmith/experiments/torch/gemma11/train.py
+++ b/blacksmith/experiments/torch/gemma11/train.py
@@ -21,6 +21,7 @@ from blacksmith.tools.torch_helpers import (
     collate_fn_for_causal_lm,
     collect_examples,
     show_examples,
+    run_decode_example_from_batch,
 )
 
 
@@ -69,6 +70,16 @@ def validate(model, val_data_loader, loss_fn, device_manager, config, logger, to
     if config.print_examples and tokenizer is not None:
         logger.info(f"\n=== Validation Examples (Random samples) ===")
         show_examples(collected_examples, tokenizer, config, logger)
+    
+    if config.run_decode_example and tokenizer is not None:
+        run_decode_example_from_batch(
+            model=model,
+            tokenizer=tokenizer,
+            batch=next(iter(val_data_loader)),
+            ignored_index=config.ignored_index,
+            device=device_manager.device,
+            logger=logger,
+        )
 
     avg_val_loss = total_val_loss / num_val_batches if num_val_batches > 0 else 0.0
     logger.info(f"Average validation loss: {avg_val_loss}")
--- a/blacksmith/tools/torch_helpers.py
+++ b/blacksmith/tools/torch_helpers.py
@@ -2,6 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 import torch
+from transformers import StaticCache
 
 
 def print_trainable_params(model):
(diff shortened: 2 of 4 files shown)


blacksmith/tools/torch_helpers.py
```diff
@@ -125,3 +126,143 @@ def collate_fn_for_causal_lm(batch):
     shifted_labels = labels[:, 1:].contiguous()
 
     return {"input_ids": input_ids, "attention_mask": attention_mask, "labels": shifted_labels}
+
+
+# ------------Decode Example From Batch------------
+def run_decode_example_from_batch(
+    model,
+    tokenizer,
+    batch,  # dict with "input_ids" (B, T) and "labels" (B, T-1)
```
mmilosevicTT: nit: Comment is not necessary, as we already say it in the docstring.

blacksmith/tools/torch_helpers.py
```diff
@@ -125,3 +126,143 @@ def collate_fn_for_causal_lm(batch):
...
+        device=device,
+        logger=logger,
+        **decode_kwargs,
+    )
+
+
+def run_decode_example(
+    model,
+    tokenizer,
+    prompt_ids,  # 1D LongTensor
```
mmilosevicTT: nit: I find this comment a bit weird.

blacksmith/tools/torch_helpers.py
```diff
@@ -125,3 +126,143 @@ def collate_fn_for_causal_lm(batch):
...
+    device,
+    logger,
+    target_ids=None,
+    max_prompt_length: int = 32,
+    max_cache_length: int = 128,
+    max_new_tokens: int = 64,
+):
+    """Run one autoregressive decode on a single prompt."""
```
mmilosevicTT: Please add a comment where from tt-xla we adapted this from, it will be easier to keep track if something changes in the main code.

blacksmith/tools/torch_helpers.py
```diff
@@ -125,3 +126,143 @@ def collate_fn_for_causal_lm(batch):
...
+
+def construct_inputs_for_decode(
+    prompt_ids,  # 1D LongTensor, len <= max_prompt_length
+    pad_token_id: int,
+    model_config,
+    max_prompt_length: int,
+    max_cache_len: int,
+):
+    """Build StaticCache inputs from a pre-tokenized prompt (batch size 1)."""
+    prompt_ids = prompt_ids[-max_prompt_length:]  # left-truncate if needed
```
mmilosevicTT: Comment is not really needed, it is obvious that this is left-side truncation.

blacksmith/tools/torch_helpers.py
```diff
@@ -125,3 +126,143 @@ def collate_fn_for_causal_lm(batch):
...
+    if target_ids is not None:
+        logger.info(f"Target: {tokenizer.decode(target_ids, skip_special_tokens=True)!r}")
+    input_args = construct_inputs_for_decode(
+        prompt_ids=prompt_ids,
+        pad_token_id=tokenizer.pad_token_id,
+        model_config=model.model.config,
+        max_prompt_length=max_prompt_length,
+        max_cache_len=max_cache_length,
+    )
+    # Transfer inputs to device
```
mmilosevicTT: nit: Keep all comments as full sentences.

mmilosevicTT (approved):
LGTM, left few nit comments

blacksmith/tools/torch_helpers.py
```diff
@@ -125,3 +126,143 @@ def collate_fn_for_causal_lm(batch):
     shifted_labels = labels[:, 1:].contiguous()
 
     return {"input_ids": input_ids, "attention_mask": attention_mask, "labels": shifted_labels}
+
+
+# ------------Decode Example From Batch------------
```
pglusacTT: Please remove this comment, it doesn't add value.

blacksmith/experiments/torch/gemma11/train.py
```diff
@@ -70,6 +71,16 @@ def validate(model, val_data_loader, loss_fn, device_manager, config, logger, to
         logger.info(f"\n=== Validation Examples (Random samples) ===")
         show_examples(collected_examples, tokenizer, config, logger)
 
+    if config.run_decode_example and tokenizer is not None:
```
mmilosevicTT, line 74: Should we do only decode example or teacher forcing examples?

mcupacTT: I would leave it for each specific script to decide - both are useful as a sanity check if the training is successful.

mcupacTT:
Resolved comments, thanks for the feedback!

PR #573, opened 2026-05-15
train_functions: remove spurious mutable=["params"] from forward_pass

Dev-X25874:
### Issue
N/A — bug found via static code review of the JAX MNIST training pipeline.

### Bug Description
`forward_pass` in `blacksmith/experiments/jax/mnist/single_chip/train_utils/train_functions.py`
called `MLP().apply({"params": params}, x, mutable=["params"])`, incorrectly marking
the `"params"` variable collection as mutable. Flax interprets this as permission to
re-initialise the parameter collection during the forward pass, so every call to
`forward_pass` effectively ran on freshly randomised weights rather than the trained
ones. As a result, gradients were computed against random weights and optimiser updates
had no effect — the model could not learn.

### What's Changed
Removed the `mutable=["params"]` argument from the `MLP().apply(...)` call and
updated the return to match the non-mutable signature (single return value, not a
2-tuple). `MLP` is a stateless `@nn.compact` module containing only `nn.Dense`
layers with no mutable state, so the correct Flax idiom is a plain read-only
`.apply({"params": params}, x)` call.

```python
# Before
logits, _ = MLP().apply({"params": params}, x, mutable=["params"])

# After
logits = MLP().apply({"params": params}, x)
```

### Environment
- **Commit/Version:** identified against `main` branch snapshot
- **Hardware:** N/A — CPU-reproducible; does not require TT hardware

### Checklist
- [x] Bug is reproducible before the fix
- [x] Fix resolves the bug
- [ ] Tests pass and updated if needed
- [x] No new bugs introduced
- [ ] Documentation updated if needed

### DIFF
--- a/blacksmith/experiments/jax/mnist/single_chip/train_utils/train_functions.py
+++ b/blacksmith/experiments/jax/mnist/single_chip/train_utils/train_functions.py
@@ -13,7 +13,7 @@ def cross_entropy(logits, labels):
 
 
 def forward_pass(params, x):
-    logits, _ = MLP().apply({"params": params}, x, mutable=["params"])
+    logits = MLP().apply({"params": params}, x)
     return logits
 
 


Dev-X25874:
Hey @vladimirjovanovicTT @umalesTT @pglusacTT @ndrakulicTT @mmilosevicTT @mcupacTT @abogdanovicTT @agobeljicTT — could you please take a look at this PR when you get a chance? 🙏

It's a small but impactful one-line fix — removing the spurious `mutable=["params"]` from the `forward_pass` call that was causing the model to train on randomly re-initialised weights every step. The change is minimal and CPU-reproducible, so should be quick to review. Happy to answer any questions!

PR #576, opened 2026-05-19
[CI] Introduce atol in loss accuracy check

umalesTT:
For small losses, near the end of training curve, `rtol` turned out to be too sensitive and correct workflows failed testing.
This PR introduces `atol` check alongside `rtol`, which will become dominant and more resilient in region of small losses, while for losses in beginning of training, `rtol` will continue to dominate.

### DIFF
--- a/tests/test_all.py
+++ b/tests/test_all.py
@@ -16,6 +16,7 @@ DEFAULT_SETUP_DICT = {
     "experiment_config": None,
     "test_config": None,
     "tolerance": 0.5,
+    "atol": 0.1,
     "timeout": 800.0,
     "skip_loss_checks": False,
     "test_checkpoint_path": None,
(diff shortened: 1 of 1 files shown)


tests/test_all.py
```diff
@@ -94,11 +95,13 @@ def check_losses(train_log_file: Path, val_log_file: Path, setup_dict: dict):
         TEST_LOGS_DIR / train_log_file,
         GOLDEN_LOGS_DIR / train_log_file,
         tolerance=setup_dict["tolerance"],
+        atol=setup_dict["atol"],
     )
```
mmilosevicTT, line 99: Are we sure this is fine to do in each step? I would expect atol to be larger in the first few steps, so keeping it too low would break the test.

I would say it is fine for now and let's see what happens to the CI in future. If it turns out to be too painful we can update. Wdyt?

mmilosevicTT (approved):
LGTM

tests/test_all.py
umalesTT: They add, so in first few steps we actually relax constraint

PR #577, opened 2026-05-19
[CI] Set Gemma 1.1 as XFAIL

umalesTT:
We are currently investigating PCC issues that appear in Gemma 1.1 training, setting test as XFAIL to unblock further uplifts until issue is solved.


mmilosevicTT (approved):
LGTM

PR #594, opened 2026-05-30
Fix Determinism In Reproducibility Manager

mmilosevicTT:
### Issue
N/A

### Bug Description
In refactor we forgot to add `.value` on fields so runs were not matching proper framework.

### What's Changed
Fixed matching patterns to call `.value` on enums and properly apply determinism in `reproducibility_manager`

### Checklist
- [x] Run single run two times.

### DIFF
--- a/blacksmith/tools/reproducibility_manager.py
+++ b/blacksmith/tools/reproducibility_manager.py
@@ -21,9 +21,9 @@ class ReproducibilityManager:
     def setup(self):
         self._setup_python()
 
-        if self.config.framework == Framework.PYTORCH:
+        if self.config.framework == Framework.PYTORCH.value:
             self._setup_pytorch()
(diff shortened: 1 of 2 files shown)


mcupacTT (approved):
LGTM!

Maybe we also add a simple log message in the setup for each framework to confirm that the corresponding framework setup was performed succesfully.
So that the user can see if the setup completed as requested.

mmilosevicTT:
We already had a message that framework did not match. It is just that it is easy to miss it when many logs are coming in. I believe that is enough. Wdyt? @mcupacTT

mcupacTT:
> We already had a message that framework did not match. It is just that it is easy to miss it when many logs are coming in. I believe that is enough. Wdyt? @mcupacTT

Yes, sure, I agree, it is enough.

PR #596, opened 2026-05-31
Fix Logging In Torch Experiments When Gradient Accumulation Is On

mmilosevicTT:
### Issue
N/A

### Bug Description
Currently in code, if number of batches in single epoch is not divisible by `gradient_accumulation_step` we just start the counter from 0 once the epoch ends. This is incorrect for logging, as `running_loss` is only reset once we enter logging phase, so we will have wrong starting point. 

### What's Changed
Move reinitialization of `running_loss` to epoch start.

### Checklist
- [x] Run experiment to confirm compilation


mcupacTT (approved):
LGTM!
Great catch!

PR #493, opened 2026-03-17
Added CPU implementation for ALE/Breakout-v5

Negoje:
### Issue
First PR for the issue https://github.com/tenstorrent/tt-blacksmith/issues/454

### Training Workload Setting

- **Model:**  BreakoutCNN — custom CNN actor-critic (~1.7M parameters); 3 conv layers (32/64/64 filters) → 512-unit FC → shared actor/critic heads
- **Strategy:** Full-model training via PPO (Proximal Policy Optimization) with GAE
- **Framework:** PyTorch
- **Task:** Reinforcement learning — Atari game playing (Breakout)
- **Dataset:** ALE/Breakout-v5 via Gymnasium with standard Atari preprocessing (84×84 grayscale, frame skip=4, frame stack=4, episodic life, fire reset, reward clipping)
- **Hardware:** CPU only (TT-XLA support not yet available)

### What's Changed
This PR showcases a PPO training workload for Atari Breakout using a standard CNN actor-critic architecture. The agent is trained using Generalized Advantage Estimation (GAE) across 8 parallel environments for 10M timesteps. Key features:

- BreakoutCNN model with orthogonal weight initialization
- RolloutBuffer with GAE computation
- Atari environment wrappers: EpisodicLifeEnv, FireResetEnv, ClipRewardEnv
- Full integration with blacksmith tooling: TrainingLogger (W&B), CheckpointManager, DeviceManager, ReproducibilityManager
- Requires `ale_py`, `gymnasium[other]` (`pip install ale_py`, `pip install gymnasium[other]`) for ALE environment support
- The CNN follows the architecture from the [DQN Nature Paper](https://storage.googleapis.com/deepmind-media/dqn/DQNNaturePaper.pdf). The same feature extractor is commonly reused for Atari environments regardless of the RL algorithm on top.

### Experiment Results
image
image

Example outputs:
- Average episodic return curve over training steps
- Loss curves: policy loss, value loss, entropy loss, approx KL, clip fraction

### Checklist
- [x] Main training workflow
- [x] CPU testing
- [ ] GPU testing
- [ ] TT device testing
- [ ] CI testing
    - [ ] Existing tests are passing
    - [ ] Register test to CI
    - [ ] Upload golden files if necessary

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1213878096103996

Issue #454, opened 2026-02-18
[$1000 Bounty] Training workload for CNN PPO on ALE/Breakout-v5


vladimirjovanovicTT:
## Add CNN PPO Single-Chip RL Training on ALE/Breakout-v5 on TT-N150 with CPU Baseline  

**Background:**  
- TT-Blacksmith is a growing collection of model demos showcasing the capabilities of AI models running on Tenstorrent hardware.  
- This bounty aims to bring up **full training** of a **CNN-based PPO** (Proximal Policy Optimization) reinforcement learning agent on the **ALE/Breakout-v5** Atari environment on a single **Tenstorrent N150** chip, faithfully reproducing the training setup in PyTorch.  
- To establish a baseline, the same training must also be run on CPU and the results compared, ensuring that metrics (e.g., episode reward curves, policy loss, value loss) align between CPU and TT-N150 runs.  

When encountering compilation or runtime issues, contributors are expected to:  
1. Delegate the issue with a minimal repro to the appropriate Tenstorrent repository (`tt-xla/tt-mlir` for compilation, `tt-metal` for runtime).  
2. Provide a **CPU fallback workaround** in their training code so training can continue while the issue is investigated.  

This bounty is an opportunity for AI developers to:  
- Contribute to Tenstorrent's model demos  
- Validate single-chip reinforcement learning training  
- Earn rewards  

---

**Requirements:**  
- **Framework:** Implementation must be in **PyTorch**.  

1. **CNN PPO Training on TT-N150:**  
   - Implement **end-to-end PPO training** with a CNN policy network on a single chip.  
   - Train on the **ALE/Breakout-v5** environment.  
   - Reproduce all training hyperparameters, environment setup, and workflow from standard PPO Atari implementations.  

2. **CPU Baseline:**  
   - Run the same training on CPU.  
   - Compare results, ensuring parity in metrics and convergence.  

3. **Fallback Mechanism:**  
   - In cases where an operation fails on TT-N150, redirect only the **minimal possible function containing that operation** to CPU.  
   - Do **not** fallback the entire training step to CPU — training must primarily execute on TT hardware.  

4. **Verification:**  
   - To confirm training is running on TT hardware, set:  
     ```bash
     export TTXLA_LOGGER_LEVEL=DEBUG
     ```  
     - Look for printed TTIR graphs (e.g., `graph module @jit_training_step`) confirming TT execution.  

5. **Koyeb Machines:**  
[...]

PR #605, opened 2026-06-06
Added TT support for ALE/Breakout-v5 (Issue #454)

Negoje:
### Issue
Second PR for the issue https://github.com/tenstorrent/tt-blacksmith/issues/454

Depends on PR #493 

### Training Workload Setting
- **Model:** BreakoutCNN — custom CNN actor-critic (~1.7M parameters); 3 conv layers (32/64/64 filters) → 512-unit FC → shared actor/critic heads
- **Strategy:** Full-model training via PPO (Proximal Policy Optimization) with GAE
- **Framework:** PyTorch
- **Task:** Reinforcement learning — Atari game playing (Breakout)
- **Dataset:** ALE/Breakout-v5 via Gymnasium with standard Atari preprocessing (84×84 grayscale, frame skip=4, frame stack=4, episodic life, fire reset, reward clipping)
- **Hardware:** Tenstorrent hardware via TT-XLA

### What's Changed
This PR extends the PPO Breakout workload from #493 with support for Tenstorrent hardware via TT-XLA. Training can now run on TT devices in addition to CPU.
For training I used ```resume_from_checkpoint: True``` and in terminal this command:
```bash
while true; do python blacksmith/experiments/torch/BOUNTIES/ppo_breakout/train.py; sleep 10; done
```
Key additions:
- TT-XLA device integration for the BreakoutCNN model
- Device-aware training loop compatible with TT hardware constraints

### Related Issues
Filed while working on this PR:
- https://github.com/tenstorrent/tt-xla/issues/5321 - index_select; workaround with matmul + onehot vector
- https://github.com/tenstorrent/tt-xla/issues/5325 - gather; workaround with matmul + onehot vector
- https://github.com/tenstorrent/tt-blacksmith/issues/609 - memory leak during training; workaround with resume from checkpoint

### Experiment Results
image
image

### Checklist
- [x] Main training workflow
- [x] CPU testing
- [ ] GPU testing
- [x] TT device testing
- [ ] CI testing
    - [ ] Existing tests are passing
    - [ ] Register test to CI
    - [ ] Upload golden files if necessary


Negoje:
During a training run at around 800k steps, the process crashed due to what appears to be a memory leak on the TT device. The root cause has not been identified yet. Can you give me a tip how to debug this?

As a workaround, I added resume from checkpoint support so training can continue after a crash. This required a small change to checkpoint manager.

blacksmith/experiments/torch/BOUNTIES/ppo_breakout/single_chip/ppo_breakout.yaml
mmilosevicTT, line 1: Please follow namings in other experiments.

blacksmith/models/torch/BOUNTIES/breakout_cnn.py
mmilosevicTT, line 1: This should not be a part of `experiment` dir, but `models`.

blacksmith/experiments/torch/BOUNTIES/ppo_breakout/single_chip/ppo_breakout.yaml
```diff
@@ -12,7 +12,10 @@ anneal_lr: True
 num_steps: 128
 gamma: 0.99
 gae_lambda: 0.95
-num_minibatches: 4
+# 16 (not 4): batch_size is num_envs * num_steps = 1024, so 16 minibatches gives a
+# minibatch of 64. A larger minibatch (e.g. 4 -> 256) makes the CNN's conv activations
+# overflow TT L1, raising "circular buffers clash with L1 buffers" during compilation.
+num_minibatches: 16
```
mmilosevicTT: I do not like mentioning 4, when we will not see it anymore. Could you just phrase it to why we need 16? Saying that it is required by the dataset/setup is just fine IMO, wdyt @ndrakulicTT ?

mmilosevicTT (commented):
Hey @Negoje,

Thanks for working on this. 

I have added few comments on the structure of the code which I forgot to add to the initial PR. Could you please address them?

blacksmith/experiments/torch/BOUNTIES/ppo_breakout/train.py
```diff
@@ -119,26 +119,44 @@ def make_vec_env(config: TrainingConfig):
...
+    h_obs = b_obs.cpu()
+    h_actions = b_actions.cpu()
+    h_log_probs = b_log_probs.cpu()
+    h_values = b_values.cpu()
+    h_adv = b_advantages.cpu()
+    h_returns = b_returns.cpu()
 
     clip_fracs = []
 
-    dataset = TensorDataset(b_obs, b_actions, b_log_probs, b_values, b_advantages, b_returns)
```
Negoje, line 129: Removed TensorDataset/DataLoader because it caused frequent recompilations on TT, slowing down training. Replaced with manual randperm shuffling.

blacksmith/experiments/torch/BOUNTIES/ppo_breakout/single_chip/ppo_breakout.yaml
```diff
@@ -12,8 +12,10 @@ anneal_lr: True
 num_steps: 128
 gamma: 0.99
 gae_lambda: 0.95
-num_minibatches: 4
-update_epochs: 4
+# Keep minibatch size small: larger ones overflow TT L1 with the CNN's conv
+# activations, raising "circular buffers clash with L1 buffers" during compilation.
```
ndrakulicTT: Is this really during compilation?
In most cases if you saw `TT_FATAL ... circual bufffers ...` it is during runtime

blacksmith/experiments/torch/BOUNTIES/ppo_breakout/train.py
```diff
@@ -119,26 +119,44 @@ def make_vec_env(config: TrainingConfig):
...
+def ppo_update(agent, optimizer, buffer, advantages, returns, config: TrainingConfig, device_manager: DeviceManager):
+    device = device_manager.device
     b_obs, b_actions, b_log_probs, b_values = buffer.flatten()
     b_advantages = advantages.reshape(-1)
     b_returns = returns.reshape(-1)
+    batch_size = b_obs.shape[0]
+
+    # Offload the batch to host (CPU) before shuffling: TT's on-device gather
+    # (index_select with a dynamic index) returns numerically wrong values that blow
+    # the loss up to inf/nan, whereas the same gather on CPU is correct.
```
ndrakulicTT: Can you create issue for this? It would be best if you can isolate it on few ops in the graph so we can file the issue to `tt-metal` directly

blacksmith/experiments/torch/BOUNTIES/ppo_breakout/breakout_rollout.py
```diff
@@ -41,6 +44,10 @@ def compute_gae(self, next_value: torch.Tensor, next_done: torch.Tensor):
                 next_val = self.values[t + 1]
             delta = self.rewards[t] + self.config.gamma * next_val * next_non_terminal - self.values[t]
             advantages[t] = last_gae = delta + self.config.gamma * self.config.gae_lambda * next_non_terminal * last_gae
+            # Sync periodically to flush the graph; otherwise the whole unrolled GAE loop
+            # accumulates into one graph and trips the "Profiler message limit exceeded" error.
+            if t % _GAE_SYNC_INTERVAL == 0:
```
ndrakulicTT: wait but isnt `num_steps = 128`, so this doesnt have an effect, or if you wanted to sync maybe we can sync at every step in main training loop

blacksmith/models/torch/BOUNTIES/breakout_cnn.py
```diff
@@ -41,7 +40,26 @@ def get_value(self, x: torch.Tensor) -> torch.Tensor:
     def get_action_and_value(self, x: torch.Tensor, action=None):
         hidden = self.network(x)
         logits = self.actor(hidden)
-        dist = Categorical(logits=logits, validate_args=False)
+        target_device = logits.device
+
+        # Offload the categorical math to CPU: on TT the softmax+entropy has poor
+        # PCC against CPU, and running it on the host fixes that. We also compute
+        # it manually rather than with torch.distributions.Categorical because the
+        # manual softmax/gather/entropy gave better precision on TT hardware.
```
ndrakulicTT: Can yo create also issue for this, with minimal repro

blacksmith/tools/checkpoints_manager.py
```diff
@@ -234,7 +234,7 @@ def load_checkpoint_path(
         if self.config.load_from_storage:
             self.storage_backend.load(checkpoint_path, checkpoint_path)
 
-        checkpoint = torch.load(checkpoint_path, map_location="cpu")
+        checkpoint = torch.load(checkpoint_path, map_location="cpu", weights_only=False)
```
ndrakulicTT: Why do you need this?

Negoje: I checked again and it turns out I was using an older checkpoint, so this is no longer needed

blacksmith/experiments/torch/BOUNTIES/ppo_breakout/breakout_rollout.py
Negoje: It does execute once when t is zero. I moved it to main

blacksmith/experiments/torch/BOUNTIES/ppo_breakout/single_chip/ppo_breakout.yaml
Negoje: Yeah it is during runtime. I rephrased it

blacksmith/experiments/torch/BOUNTIES/ppo_breakout/train.py
Negoje: https://github.com/tenstorrent/tt-xla/issues/5321

blacksmith/models/torch/BOUNTIES/breakout_cnn.py
Negoje: https://github.com/tenstorrent/tt-xla/issues/5325

blacksmith/experiments/torch/BOUNTIES/ppo_breakout/train.py
```diff
@@ -269,10 +287,13 @@ def train(
             with torch.no_grad():
                 next_value = agent.get_value(obs).flatten()
             advantages, returns = buffer.compute_gae(next_value, done)
+            # Empirically faster on TT.
```
pglusacTT: Please remove this comment.

blacksmith/experiments/torch/BOUNTIES/ppo_breakout/single_chip/ppo_breakout.yaml
```diff
@@ -12,8 +12,10 @@ anneal_lr: True
 num_steps: 128
 gamma: 0.99
 gae_lambda: 0.95
-num_minibatches: 4
-update_epochs: 4
+# Keep minibatch size small: larger ones overflow TT L1 with the CNN's conv
+# activations, raising a "circular buffers clash with L1 buffers" at runtime.
```
pglusacTT: No need for this comment. This is something that is generally explained in the PR description/comment.

blacksmith/experiments/torch/BOUNTIES/ppo_breakout/train.py
```diff
@@ -163,11 +169,23 @@ def ppo_update(agent, optimizer, buffer, advantages, returns, config: TrainingCo
 
             optimizer.zero_grad()
             loss.backward()
+            if config.use_tt:
+                torch_xla.sync(wait=True)
             torch.nn.utils.clip_grad_norm_(agent.parameters(), config.max_grad_norm)
-            optimizer.step()
+            device_manager.optimizer_step(optimizer)
+
+    # Move all metrics to CPU for logging to avoid XLA sync issues.
```
pglusacTT: This doesn't fully match what the code actually does. `sync` will compute the results but not actually move the tensors back to cpu. IMO just remove this comment, it doesn't need any explanation.

blacksmith/models/torch/BOUNTIES/breakout_cnn.py
```diff
@@ -41,7 +40,19 @@ def get_value(self, x: torch.Tensor) -> torch.Tensor:
...
+        log_probs = torch.log_softmax(logits, dim=-1)
+        probs = log_probs.exp()
+
         if action is None:
-            action = dist.sample()
-        return action, dist.log_prob(action), dist.entropy(), self.critic(hidden)
+            # Sample on device with the Gumbel-max trick, not torch.multinomial: multinomial
+            # adds a random node to the graph that re-draws each execution, desyncing the
+            # action from its log_prob.
+            u = torch.rand(log_probs.shape, device="cpu").to(target_device)
```
pglusacTT, line 52: This is a bit suspicious to me. Why is using random instead of multinomial any different? Can you provide the IR snippets and explain this in more detail?

pglusacTT:
Why run with `while true; do python blacksmith/experiments/torch/BOUNTIES/ppo_breakout/train.py; sleep 10; done` ? Does a singular run converge? Please elaborate.

blacksmith/models/torch/BOUNTIES/breakout_cnn.py
Negoje: The problem is that with on-device multinomial, the RNG draw is a node in the lazy graph. `action` and `log_prob` both depend on that node, but get materialized in two separate executions. `action` materializes immediately in `collect_rollout` (via `.cpu()`) to step the env. `log_prob` stays lazy and materializes later. Since `action` was never stored as a value, that second execution has to recompute `action` from scratch, which re-runs the RNG and draws a different value, so `log_prob`'s gather ends up on a different `action` than the one we stepped with.

So the idea is more to avoid random node on the device and just pass the value which is why we calculate on cpu the random value. And the Gumbel-max is used just to keep everything done on tt device instead of moving the whole calculation to the cpu.

Negoje:
Hey @pglusacTT - We use that command because there is currently a memory leak on the TT device (issue #609), which causes the run to crash around roughly 1M steps. The command is just a convenience so training automatically restarts from where it left off instead of requiring manual intervention every time it crashes.

blacksmith/experiments/torch/BOUNTIES/ppo_breakout/single_chip/ppo_breakout.yaml
```diff
@@ -12,8 +12,10 @@ anneal_lr: True
...
+num_minibatches: 16
```
Negoje, line 15: Minibatch size should be small because larger ones overflow TT L1 with the CNN's conv activations, raising a "circular buffers clash with L1 buffers" at runtime.

PR #623, opened 2026-06-17
[Trainer] Add Initial Trainer Skeleton

### LINKED ISSUE #616: Add Trainer Class Skeleton
### Is your feature request related to a problem? Please describe

As part of the need to unify running of experiments we need to add a proper Trainer class for TT-Blacksmith experiments.

### Describe the solution you'd like

This issue is related to adding some initial basic logic for how trainer class should look. This includes all non training strategy specific features like callbacks and config passing.

mmilosevicTT:
### Issue
Closes https://github.com/tenstorrent/tt-blacksmith/issues/616

### Problem description
This PR is adding a basic structure of trainer class. It is meant to be a first base PR on which we can build additional features and extend it's use.

### What's Changed
This PR adds:
- Base `Trainer` class.
- Base `Callbacks` class for running additional code during training.
- Skeleton for specific strategy implementations.
- Tests for callback handling.

### Checklist
- [x] New/Existing tests provide coverage for changes


blacksmith/tools/trainer/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+from blacksmith.tools.trainer.callback import Callback
+from blacksmith.tools.trainer.callbacks_handler import CallbackHandler
+from blacksmith.tools.trainer.trainer import Trainer
+from blacksmith.tools.trainer.utils import normalize_callbacks
+
+__all__ = [
+    "Trainer",
+    "Callback",
+    "CallbackHandler",
+    "normalize_callbacks",
```
pglusacTT: Does `normalize_callbacks` need to be exposed outside of the trainer? Sounds like an internal method.

blacksmith/tools/trainer/callbacks_handler.py
```diff
@@ -0,0 +1,22 @@
...
+from collections.abc import Sequence
+
+from blacksmith.tools.trainer.callback import Callback
+
+
+class CallbackHandler:
+    def __init__(self, callbacks: Sequence[Callback]):
+        self.callbacks = list(callbacks)
+
+    def call(self, method_name: str, *args, **kwargs) -> None:
```
pglusacTT: Why not use `__call__` instead?

blacksmith/tools/trainer/callbacks_handler.py
```diff
@@ -0,0 +1,22 @@
...
+        if method_name.endswith("_start"):
+            ordered_callbacks = self.callbacks
+        elif method_name.endswith("_end"):
+            ordered_callbacks = reversed(self.callbacks)
```
pglusacTT: Synced offline; good feature idea but let's see if we have a use case for it. The code becomes simpler without it.

pglusacTT (commented):
Thanks!

blacksmith/tools/trainer/strategies/__init__.py
ndrakulicTT, line 1: Is idea to support the trainer for both jax and torch?
If not maybe better place would be inside some torch specific folder

blacksmith/tools/trainer/trainer.py
```diff
@@ -0,0 +1,51 @@
...
+        self,
+        callbacks: Union[Callback, Sequence[Callback], None] = None,
+    ):
+        self.config = None
+        self.callback_handler = CallbackHandler(normalize_callbacks(callbacks))
+
+    @abstractmethod
+    def setup(
+        self,
+        config: str | None = None,
```
ndrakulicTT: Why is config str? Is idea to load yaml from that Path? if yes would prefer `Path` object.
Also maybe is better to pass already parsed config object so that we have option to run the training with in memory objects, instead of having round trip to a file

mmilosevicTT: Yeah, I do not have strong preference for this. I just wanted for the first version to keep it simple and load config from file. Then later I would like to add some way of defaulting to experiments in the repo if we have them and otherwise require the caller to pass all required parameters.

Would you like me to switch to in memory object?

blacksmith/tools/trainer/strategies/__init__.py
mmilosevicTT: I am not excluding the idea of jax, but initial version will be torch only. Should we keep it simple until we decide to add Jax one?

blacksmith/tools/trainer/__init__.py
mmilosevicTT: Yeah, you are right, removed this import.

blacksmith/tools/trainer/callbacks_handler.py
mmilosevicTT: Changed. Good catch.

blacksmith/tools/trainer/callbacks_handler.py
mmilosevicTT: @pglusacTT lmk if current version works for you.

blacksmith/tools/trainer/trainer.py
pglusacTT: I strongly agree with using `Path` or parsed config object.

tests/trainer/test_callbacks_handler.py
```diff
@@ -0,0 +1,62 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
```
pglusacTT, line 1: When will these tests be invoked, do we have them in CI?

tests/trainer/test_callbacks_handler.py
```diff
@@ -0,0 +1,62 @@
...
+    trainer = object()
+    handler = CallbackHandler(
+        trainer,
+        [
+            RecordingCallback("first", events),
+            RecordingCallback("second", events),
+        ],
+    )
+
+    handler("on_train_batch_start", batch=None)
```
pglusacTT, line 41: I am kinda worried about using plain strings for this, we could potentially have events that are never called because we had a typo or something. What do you think about using an enum instead?

tests/trainer/test_callbacks_handler.py
mmilosevicTT: Good catch, I would prefer doing CI addition in a separate PR once we have some working example of training through trainer. Wdyt?

Can open a sub-task for it to keep record of it.

tests/trainer/test_callbacks_handler.py
mmilosevicTT: Uh, this is a tricky one.

So far it is not possible that typo is not noticed, handler would error out in `getattr(callback, event)(self.trainer, *args, **kwargs)`, since it would not find the method. Also this way we keep it easy to extend callback for specific calls for some callbacks that might appear only in specific trainers.

On the other side enum is easier to track and we can always extend the functionality with separate enums that have additional fields.

So far I can keep it with enum, but might need to adapt in future if we see that extensions become painful.

blacksmith/tools/trainer/trainer.py
mmilosevicTT: Moved to passing TrainingConfig for now.

tests/trainer/test_callbacks_handler.py
pglusacTT: Sounds good to me!

mmilosevicTT: Added TODO with the issue.

blacksmith/tools/trainer/strategies/__init__.py
mmilosevicTT: Keeping project structure as is for now.

tests/trainer/test_callbacks_handler.py
mmilosevicTT: Synced offline. We looked at pytorch lightning and they keep it simple with strings, so for now we will keep it that way.

pglusacTT (approved):
Great!

Issue #626, opened 2026-06-18
Enable Trace Mechanism for Training Scripts


pglusacTT:
TT-XLA exposes the `enable_trace` compilation option which should yield performance boosts. 

Trace mechanism is described in tt-metal repo, [found here](https://github.com/tenstorrent/tt-metal/blob/c7867a1a8700c89def0a0eacaea669dab2a19945/tech_reports/AdvancedPerformanceOptimizationsForModels/AdvancedPerformanceOptimizationsForModels.md#1-metal-trace).

Issue #625, opened 2026-06-18
DRAM Usage Regression


pglusacTT:
### Describe the bug

LLaMA 3.1 8B used to be able to run BS=1 SEQ=1024 but now it OOMs. BS=1 SEQ=512 still works.

### Steps to reproduce the issue

Run training script with described parameters.

Issue #615, opened 2026-06-17
Trainer Class for TT-Blacksmith Experiments


mmilosevicTT:
### Is your feature request related to a problem? Please describe

As part of the experiments running unification it is necessary to add a proper `Trainer` class to TT-Blacksmith.

### Describe the solution you'd like

As this is an extensive change, it is split into several tasks that incrementally enable discussion and development.

PR #627, opened 2026-06-18
[PERF] Join FWD + BWD and Sync Tensors Precisely

pglusacTT:
### Issue
N/A

---

### Problem description

By default, tt-torch backend inside tt-xla does some variant of `torch_xla.sync()` [call during model execution](https://github.com/tenstorrent/tt-xla/blob/a0ba4ca3266475f441f3d326d612d5659c5845ae/python_package/tt_torch/backend/backend.py#L272). 
[Recent PR](https://github.com/tenstorrent/tt-xla/pull/5235) on the tt-xla repo allows for purely lazy execution. 

This allows us to lazily join FWD & BWD graphs and to be surgical about which tensors we should sync. 

This yields performance improvements since we avoid redundant computation.

---

### Perf Improvement noticed on Mixa's QB2:

**LLaMA 3.1 8B, BS=1 SEQ=512**

| Old | New | Improvement |
| --- | --- | --- |
5.13s | 3.74s | 27.2%

**LLaMA 70B L=40, BS=1 SEQ=512**
| Old | New | Improvement |
| --- | --- | --- |
2.156s | 1.858s | 13.82%

Training curve and loss values don't seem to change.

---

### Notes:

- Noticed a memory regression described [here](https://github.com/tenstorrent/tt-blacksmith/issues/625).
- The LLaMA 70B numbers look a bit suspicious. Will look into during trace effort. https://github.com/tenstorrent/tt-blacksmith/issues/626
- AOT Autograd will require more work on this since by default it joins them which is undesirable since tensors remain persistent. There are a few ways to tackle this, currently prototyping. 

---

### What's Changed
- Change LLaMA script to include performance improved calls.
- Make model compilation configurable.

### Checklist

### DIFF
--- a/blacksmith/experiments/torch/llama/xla/train.py
+++ b/blacksmith/experiments/torch/llama/xla/train.py
@@ -92,7 +92,6 @@ def compute_loss(batch, model, loss_fn, gradient_accumulation_steps):
     logits = output.logits
     shift_logits = logits[:, :-1, :].contiguous()
     loss = loss_fn(shift_logits, batch["expected_output"], batch["labels_mask"])
-    # Scale by the accumulation steps so the effective-batch gradient is the mean.
     return loss / gradient_accumulation_steps
 
 
@@ -215,8 +214,6 @@ def train(
                         logger.info(f"Step {global_step} e2e time: {step_elapsed:.3f}s")
 
                     if global_step % config.steps_freq == 0:
-                        # loss_ is already scaled by gradient_accumulation_steps, so running_loss
-                        # holds the summed per-step mean loss; divide by steps_freq for the window mean.
                         avg_loss = running_loss / config.steps_freq
                         logger.log_metrics({"train/loss": avg_loss}, commit=False, step=global_step)
                         running_loss = 0.0
--- a/blacksmith/experiments/torch/llama/xla/train.py
+++ b/blacksmith/experiments/torch/llama/xla/train.py
@@ -87,14 +87,11 @@ def validate(model, val_data_loader, loss_fn, logger, device, config, tokenizer=
     return avg_val_loss
 
 
-def training_step_inner(batch, model, loss_fn, gradient_accumulation_steps):
+def compute_loss(batch, model, loss_fn):
     output = model(input_ids=batch["input_ids"], attention_mask=batch["attention_mask"])
     logits = output.logits
     shift_logits = logits[:, :-1, :].contiguous()
-    loss = loss_fn(shift_logits, batch["expected_output"], batch["labels_mask"])
-    # Scale loss by number of accumulation steps to get correct effective batch size.
-    scaled_loss = loss / gradient_accumulation_steps
-    return scaled_loss
+    return loss_fn(shift_logits, batch["expected_output"], batch["labels_mask"])
 
 
 def train(
@@ -130,9 +127,17 @@ def train(
 
     tokenizer = train_dataset.tokenizer
 
-    compile_options = {"tt_enable_torch_fx_fusion_pass": False, "tt_legacy_compile": True, "tt_lazy_execution": True}
-    train_fn = torch.compile(training_step_inner, backend="tt", options=compile_options)
-    model_compiled = torch.compile(model, backend="tt", options=compile_options)
+    if config.use_tt:
+        compile_options = {
+            "tt_enable_torch_fx_fusion_pass": False,
+            "tt_legacy_compile": True,
+            "tt_lazy_execution": True,
+        }
+        compute_loss_fn = torch.compile(compute_loss, backend="tt", options=compile_options)
+        eval_model = torch.compile(model, backend="tt", options=compile_options)
+    else:
+        compute_loss_fn = compute_loss
+        eval_model = model
 
     global_step = 0
 
@@ -140,7 +145,7 @@ def train(
         # Initial validation
         model.eval()
         val_loss = validate(
-            model_compiled,
+            eval_model,
             eval_dataloader,
             cross_entropy_loss,
             logger,
@@ -183,15 +188,17 @@ def train(
                 # Shard model if tensor parallelism is used.
                 device_manager.shard_model(model)
 
-                # Forward (compiled) + backward (traced lazily into the same graph).
-                loss_ = train_fn(batch, model, cross_entropy_loss, config.gradient_accumulation_steps)
-                loss_.backward()
+                loss_ = compute_loss_fn(batch, model, cross_entropy_loss)
+                # Scale by the accumulation steps so the effective-batch gradient is the mean.
+                scaled_loss = loss_ / config.gradient_accumulation_steps
+                scaled_loss.backward()
 
-                # Sync only gradients and loss.
-                tensors_to_sync = [loss_] + [p.grad for p in trainable_params if p.grad is not None]
-                devices = list({t.device.type for t in tensors_to_sync})
-                torch_xla._XLAC._xla_sync_multi(tensors_to_sync, devices, wait=True)
-                torch_xla._XLAC._clear_pending_irs(torch_xla._XLAC._xla_get_default_device())
+                if config.use_tt:
+                    tensors_to_sync = [loss_] + [p.grad for p in trainable_params if p.grad is not None]
+                    devices = list({t.device.type for t in tensors_to_sync})
+                    torch_xla._XLAC._xla_sync_multi(tensors_to_sync, devices, wait=True)
+                    # Optimizer will do unnecessary recompute if we don't clear the pending IRs after syncing the loss and grads.
+                    torch_xla._XLAC._clear_pending_irs(torch_xla._XLAC._xla_get_default_device())
 
                 running_loss += loss_.item()
                 accumulation_step += 1
@@ -216,7 +223,7 @@ def train(
                     if global_step % config.val_steps_freq == 0:
                         model.eval()
                         val_loss = validate(
-                            model_compiled,
+                            eval_model,
                             eval_dataloader,
                             cross_entropy_loss,
                             logger,
--- a/blacksmith/experiments/torch/llama/xla/train.py
+++ b/blacksmith/experiments/torch/llama/xla/train.py
@@ -87,11 +87,6 @@ def validate(model, val_data_loader, loss_fn, logger, device, config, tokenizer=
     return avg_val_loss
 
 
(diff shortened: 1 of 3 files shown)


blacksmith/experiments/torch/llama/xla/train.py
```diff
@@ -135,13 +130,17 @@ def train(
 
     tokenizer = train_dataset.tokenizer
 
+    compile_options = {"tt_enable_torch_fx_fusion_pass": False, "tt_legacy_compile": True, "tt_lazy_execution": True}
```
ndrakulicTT: Why not enable `tt_lazy_execution` by default in `blacksmith/models/torch/huggingface/hf_models.py`

blacksmith/experiments/torch/llama/xla/train.py
```diff
@@ -184,11 +183,15 @@ def train(
...
+                loss_ = train_fn(batch, model, cross_entropy_loss, config.gradient_accumulation_steps)
+                loss_.backward()
 
-                if config.use_tt:
-                    torch_xla.sync(wait=True)
+                # Sync only gradients and loss.
+                tensors_to_sync = [loss_] + [p.grad for p in trainable_params if p.grad is not None]
+                devices = list({t.device.type for t in tensors_to_sync})
+                torch_xla._XLAC._xla_sync_multi(tensors_to_sync, devices, wait=True)
+                torch_xla._XLAC._clear_pending_irs(torch_xla._XLAC._xla_get_default_device())
```
ndrakulicTT: For my info, what does `_clear_pending_irs` do?

blacksmith/experiments/torch/llama/xla/train.py
```diff
@@ -184,11 +183,15 @@ def train(
...
 
                 running_loss += loss_.item()
```
abogdanovicTT, line 202: We return `scaled_loss` and add that to `running_loss`, does it mean `avg_loss` will be smaller?

blacksmith/experiments/torch/llama/xla/train.py
agobeljicTT: maybe localize the change first?

blacksmith/experiments/torch/llama/xla/train.py
```diff
@@ -87,20 +87,14 @@ def validate(model, val_data_loader, loss_fn, logger, device, config, tokenizer=
...
 def training_step_inner(batch, model, loss_fn, gradient_accumulation_steps):
     output = model(input_ids=batch["input_ids"], attention_mask=batch["attention_mask"])
     logits = output.logits
     shift_logits = logits[:, :-1, :].contiguous()
     loss = loss_fn(shift_logits, batch["expected_output"], batch["labels_mask"])
     # Scale loss by number of accumulation steps to get correct effective batch size.
     scaled_loss = loss / gradient_accumulation_steps
-    scaled_loss.backward()
-    return loss.detach()
+    return scaled_loss
```
agobeljicTT: this function now doesn't perform backward, making its name a bit incorrect
perhaps something like forward pass

blacksmith/experiments/torch/llama/xla/train.py
agobeljicTT: guard with config.use_tt

blacksmith/experiments/torch/llama/xla/train.py
agobeljicTT: yep, this should be changed

blacksmith/experiments/torch/llama/xla/train.py
pglusacTT: Good point!

blacksmith/experiments/torch/llama/xla/train.py
mmilosevicTT: This will break GPU run, since we are using "tt" backend.

blacksmith/experiments/torch/llama/xla/train.py
```diff
@@ -135,13 +130,17 @@ def train(
 
     tokenizer = train_dataset.tokenizer
 
+    compile_options = {"tt_enable_torch_fx_fusion_pass": False, "tt_legacy_compile": True, "tt_lazy_execution": True}
+    train_fn = torch.compile(training_step_inner, backend="tt", options=compile_options)
+    model_compiled = torch.compile(model, backend="tt", options=compile_options)
```
mmilosevicTT: I am not a fan of this going here without a guard. If someone flips `compile_model` in `get_model` we will error out here.

blacksmith/experiments/torch/llama/xla/train.py
pglusacTT: Yes, good catch

blacksmith/experiments/torch/llama/xla/train.py
pglusacTT: Agreed with @agobeljicTT . It is out of scope for the effort but I'd push for it in https://github.com/tenstorrent/tt-blacksmith/issues/615.

mmilosevicTT: As we discussed offline. I agree with @ndrakulicTT , but I would just check that pushing compilation of model to `get_model` works without this refactor. We can always use different flag `lazy_execution` for options where we want to enable this flag and not keep it that full compilation goes out of `get_model`.

blacksmith/experiments/torch/llama/xla/train.py
pglusacTT: There was redundant recomputation of gradients done in the optimizer IR (still not 100% sure why, I think it is related to some gradient view that is not synced). After backward, optimizer will do a `mark_step` which will compute any tensors that are left uncomputed.  With `_clear_pending_irs` you effectively just disregard any compute that isn't related to the optimizer (like view materialization etc). 

I double checked the grad values and they are the same in the old path and in the new one, the training curve follows expected values. Tried accumulation as well, everything is ok.

blacksmith/experiments/torch/llama/xla/train.py
```diff
@@ -135,13 +127,25 @@ def train(
...
     tokenizer = train_dataset.tokenizer
 
+    if config.use_tt:
+        compile_options = {
+            "tt_enable_torch_fx_fusion_pass": False,
+            "tt_legacy_compile": True,
+            "tt_lazy_execution": True,
+        }
+        compute_loss_fn = torch.compile(compute_loss, backend="tt", options=compile_options)
+        eval_model = torch.compile(model, backend="tt", options=compile_options)
```
mmilosevicTT: Just out of curiosity, I see you noticed DRAM regression, but have you checked the host memory usages for this, as well? Does torch use same underlying model for both compilations or does it create a copy in memory? 

I am worried for this as we already have opened issues for TT -> CPU movements of models where memory leaks constantly in each loop (https://github.com/tenstorrent/tt-xla/issues/5292).

blacksmith/experiments/torch/llama/xla/train.py
```diff
@@ -184,11 +188,17 @@ def train(
                 # Shard model if tensor parallelism is used.
                 device_manager.shard_model(model)
 
-                # Training step.
-                loss_ = training_step_inner(batch, model, cross_entropy_loss, config.gradient_accumulation_steps)
+                loss_ = compute_loss_fn(batch, model, cross_entropy_loss)
+                # Scale by the accumulation steps so the effective-batch gradient is the mean.
+                scaled_loss = loss_ / config.gradient_accumulation_steps
```
mmilosevicTT: Why don't we leave this in `compute_loss_fn` and just remove scaling in logging and running loss accumulation?

blacksmith/experiments/torch/llama/xla/train.py
```diff
@@ -184,11 +188,17 @@ def train(
...
+                scaled_loss.backward()
 
                 if config.use_tt:
-                    torch_xla.sync(wait=True)
+                    tensors_to_sync = [loss_] + [p.grad for p in trainable_params if p.grad is not None]
```
mmilosevicTT, line 196: Should't we use `scaled_loss` here?

blacksmith/experiments/torch/llama/xla/train.py
```diff
@@ -135,13 +127,25 @@ def train(
...
+    if config.use_tt:
+        compile_options = {
+            "tt_enable_torch_fx_fusion_pass": False,
+            "tt_legacy_compile": True,
+            "tt_lazy_execution": True,
+        }
+        compute_loss_fn = torch.compile(compute_loss, backend="tt", options=compile_options)
+        eval_model = torch.compile(model, backend="tt", options=compile_options)
+    else:
+        compute_loss_fn = compute_loss
```
mmilosevicTT: nit: Not a fan of this naming `compute_loss_fn`. Generally compute is redundant here if you ask me, `loss_fn` itself implies calculation.

mmilosevicTT: Maybe even rename `compute_loss` to something like `scoped_loss_fn` or even `unscaled_loss_fn` if you decide not to move scaling to the loop.

mmilosevicTT (commented):
Looks good in general.

I only have 2 concerns in line 137 and 197. Otherwise nits 😄

blacksmith/experiments/torch/llama/xla/train.py
pglusacTT: Renamed: `compute_loss_fn` -> `loss_step` and `compute_loss` -> `scoped_loss_fn` (kept the `scoped` part since the function still scopes the large logits tensor locally).

blacksmith/experiments/torch/llama/xla/train.py
pglusacTT: Good question. `torch.compile` does not copy the model; both `loss_step` (which receives `model` as an argument) and `eval_model = torch.compile(model)` reference the same `nn.Module` instance and the same parameter tensors (`eval_model._orig_mod is model`), so there is no duplicate set of weights on host. The two compiles only produce two cached graphs/executables (forward-only for validation vs. forward+loss+backward for training), which costs some host memory for the compiled artifacts but not a second copy of the weights.

mmilosevicTT (approved):
LGTM

nit: As we discussed offline, I realise that my previous comment is a bit off. `loss_fn` should regard to just logits and loss. Could you revert to `compute_loss`.

PR #630, opened 2026-06-19
Add support for enable_trace 

abogdanovicTT:
**Issue**
N/A

**Problem description**

TT-XLA exposes the `enable_trace` compilation option, which should give performance improvements by reducing per-step host overhead. Right now, tt-blacksmith training scripts do not have a clean way to enable trace and test it in workloads.

This PR adds support for trace-enabled execution in tt-blacksmith Llama training experiments so we can run and compare trace vs non-trace more easily.

**What's Changed**

- Add support for an `enable_trace` option in tt-blacksmith training experimets.
- Add the trace option through the relevant config/runtime path.
- Keep the current behavior unchanged when trace is disabled.


blacksmith/tools/device_manager.py
```diff
@@ -51,6 +51,10 @@ def _setup_tt_environment(self):
             os.environ["DISABLE_NUMERIC_CC_TOKEN"] = "1"
             xr.use_spmd()
 
+        # Additional setup for DRAM region for runtime trace (before device initialization).
+        if hasattr(self.config, "enable_trace") and self.config.enable_trace is True:
```
mmilosevicTT: `enable_trace` is always set. You are not using Optional value in the yaml.
```suggestion
        if self.config.enable_trace:
```

mmilosevicTT (approved):
LGTM

blacksmith/tools/device_manager.py
abogdanovicTT: We should check where `TrainingConfig` template is used in the experiment configs. For the current setup, I think we could keep this.

Issue #609, opened 2026-06-11
Memory leak while training PPO ALE/Breakout-v5


Negoje:
### Describe the bug

While building PPO training (ALE/Breakout-v5) on TT hardware for bounty #454, host memory leaks steadily throughout training. The chart below shows process memory available during CPU vs TT training — on TT it declines linearly for the entire run until it would crash. CPU training is flat.

Current workaround is to resume from a checkpoint periodically, which resets the leak.

<img width="400" height="238" alt="Image" src="https://github.com/user-attachments/assets/f0e41eff-2ca2-4e70-b706-98fd7086396f" />

### Steps to reproduce the issue

1. Check PR #605 
2. Run: python blacksmith/experiments/torch/BOUNTIES/ppo_breakout/test_breakout_ppo_training.py
3. Wait for the training to do about 800k-900k

terminate called without an active exception                                                                                                                                                             [0/1833]
[fbc167003b86:1640037] *** Process received signal ***                                                                                                                                                           
[fbc167003b86:1640037] Signal: Aborted (6)                                                                                                                                                                       
[fbc167003b86:1640037] Signal code:  (-6)                                                                                                                                                                        
[fbc167003b86:1640037] [ 0] /lib/x86_64-linux-gnu/libc.so.6(+0x45330)[0x7cab90444330]                                                                                                                            
[fbc167003b86:1640037] [ 1] /lib/x86_64-linux-gnu/libc.so.6(syscall+0x1d)[0x7cab9052628d]                                                                                                                        
[fbc167003b86:1640037] [ 2] /home/ttuser/fnegojevic/tt-blacksmith/env/xla_env/lib/python3.12/site-packages/pjrt_plugin_tt/lib64/libTTMLIRCompiler.so(+0x9688944)[0x7caafce88944]                                 
[fbc167003b86:1640037] [ 3] /lib/x86_64-linux-gnu/libc.so.6(+0x45330)[0x7cab90444330]                                                                                                                            
[fbc167003b86:1640037] [ 4] /lib/x86_64-linux-gnu/libc.so.6(pthread_kill+0x11c)[0x7cab9049db2c]                                                                                                                  
[fbc167003b86:1640037] [ 5] /lib/x86_64-linux-gnu/libc.so.6(gsignal+0x1e)[0x7cab9044427e]                                                                                                                        
[...]

PR #634, opened 2026-06-21
fix(#609): Memory leak while training PPO ALE/Breakout-v5

Notime02:
## Automatic Fix

The memory leak appears to be related to TT hardware training where tensors and buffers aren't being properly released. The fix involves adding explicit cleanup of distributed buffers and tensors after each training step.

---
*This PR was created automatically by Bounty Bot.*


mcupacTT:
Closing PR as we suspect it is malicious for the repo.

PR #635, opened 2026-06-21
fix(#609): Memory leak while training PPO ALE/Breakout-v5

Notime02:
## Fix

The memory leak issue during PPO training on TT hardware is likely due to improper resource management in the training loop. To mitigate this, I added explicit calls to free up resources after each training iteration. This should help in managing memory more effectively and prevent the linear decline in available memory.

---
*Otomatik oluşturulmuştur.*


mmilosevicTT:
Closing PR as we suspect it is malicious for the repo.

PR #640, opened 2026-06-24
Reduce batch size for LLaMA 3.1 8B Test

pglusacTT:
### Problem Description

Uplift is [failing](https://github.com/tenstorrent/tt-blacksmith/actions/runs/28081486327/job/83137253197?pr=638) due to OOM issues caused by https://github.com/tenstorrent/tt-blacksmith/pull/627.

Currently exploring and will resolve fully the AOT Autograd effort. Just reducing the batch size for now.

Probable causes:
- Joining FWD and BWD forces a certain level of recomputation that might have unintentionally saved up some memory.
- Noticed in tt-swiss reports that startup DRAM memory is higher than previously working commit. Might have to do with const eval across both fwd in bwd.


tests/configs/tt-llama_3_1_8b-sst2-data_tensor_parallel-n300-llmbox.yaml
```diff
@@ -0,0 +1,2 @@
+test_config:
```
mmilosevicTT, line 1: Ah, I am really not a fan of this not accounting for gradient_accumulation, but that might be a good discussion for CI improvements later.

mmilosevicTT (approved):
LGTM

Left just one note about thing we might want to discuss in future 😄

tests/configs/tt-llama_3_1_8b-sst2-data_tensor_parallel-n300-llmbox.yaml
pglusacTT: Yup, let's discuss this later.

PR #652, opened 2026-06-29
ci: add OpenSearch indexing step to docs deploy workflow

ctr-fmamedov-crypto:
## Summary

- `call-build-docs.yml`: adds `environment: github-pages`, OpenSearch env vars (`DOC_SITE_BASE_URL: https://docs.tenstorrent.com/tt-blacksmith`), `fetch-depth: 0`, `fetch-tags: true`, warn + index steps with `continue-on-error: true`
- `scripts/index_remote_search.py`: generic HTML crawler, reads all config from env vars

Indexing failures never block page deploys. Requires secret `DOCS_SEARCH_INGEST_API_KEY_DOCS_TENSTORRENT` to be set in repo settings.

## Merge order

Independent — can be merged at any time.

## Test plan

- [ ] Docs deploy succeeds when secret is absent (warn step fires, index skipped)
- [ ] Search results appear after deploy with secret set

### DIFF
--- a/.github/workflows/call-build-docs.yml
+++ b/.github/workflows/call-build-docs.yml
@@ -14,10 +14,22 @@ jobs:
   build:
     runs-on:
       - ubuntu-latest
+    environment: github-pages
+    env:
+      SEARCH_API_BASE: https://csl860x2oj.execute-api.us-east-2.amazonaws.com/prod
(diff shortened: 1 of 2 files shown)


mmilosevicTT (approved):
LGTM

PR #660, opened 2026-07-06
[PERF] Disable const-eval for LLaMA 3.1 8B single-chip LoRA

abogdanovicTT:
### Issue

N/A

### Problem description

By default const-eval is enabled in tt-blacksmith. During compilation this folds every subgraph whose inputs are graph constants (weight-preprocessing: bf16 - f32 casts, RMSNorm / LoRA weight prep) into CPU-hoisted functions. 
In the compiled IR for LLama 3.1 8B that produces 350 `main_const_eval_*` functions and 270 `ttir.cpu_hoist_call`s into 33 `cpu_hoisted_const_eval_*` CPU functions.

This PR exposes an `enable_const_eval` compile option (default `True`) and sets it to `False` for the LLaMA 3.1 8B single-chip LoRA config, keeping those ops on-device.

### Perf improvement (LLaMA 3.1 8B, BS=1, SEQ=512, single chip)

| Old (const-eval on) | New (const-eval off) | Improvement |
| ------------------- | -------------------- | ----------- |
| 4959s              | 1448s               | 29.2%       |

### What's Changed

- Add `enable_const_eval: bool = True` to `TrainingConfig` (`experiments/torch/llama/configs.py` and `tools/templates/configs.py`).
- Forward `enable_const_eval` into `torch_xla.set_custom_compile_options` in `experiments/torch/llama/xla/train.py`.
- Set `enable_const_eval: False` in `xla/lora/single_chip/llama_3_1_8b_sst2.yaml`.

### Notes

- Default behavior is unchanged (`enable_const_eval=True`); only Llama 8B single-chip LoRA config opts out.
- The same yaml also sets `max_length: 512` and `batch_size: 1` to match the BS=1/SEQ=512 configuration used for these PERF measurements.

### DIFF
--- a/blacksmith/experiments/torch/llama/xla/lora/single_chip/llama_3_1_8b_sst2.yaml
+++ b/blacksmith/experiments/torch/llama/xla/lora/single_chip/llama_3_1_8b_sst2.yaml
@@ -3,13 +3,13 @@ dataset_id: "sst2"
 
 # Model settings
 model_name: "meta-llama/Llama-3.1-8B"
-max_length: 512
+max_length: 128
 dtype: "torch.bfloat16"
 
 # Training hyperparameters
 training_model_type: "lora"
 learning_rate: 1e-4
-batch_size: 1
+batch_size: 8
 num_epochs: 1
 ignored_index: -100
 
--- a/blacksmith/experiments/torch/llama/xla/lora/single_chip/llama_3_1_8b_sst2.yaml
+++ b/blacksmith/experiments/torch/llama/xla/lora/single_chip/llama_3_1_8b_sst2.yaml
@@ -3,13 +3,13 @@ dataset_id: "sst2"
 
 # Model settings
 model_name: "meta-llama/Llama-3.1-8B"
-max_length: 128
+max_length: 512
 dtype: "torch.bfloat16"
 
 # Training hyperparameters
 training_model_type: "lora"
 learning_rate: 1e-4
-batch_size: 8
+batch_size: 1
 num_epochs: 1
 ignored_index: -100
 
@@ -59,3 +59,4 @@ use_tt: True
 measure_e2e_time: False
 enable_trace: False
 trace_region_size: 1000000000 # DRAM region size (bytes) for runtime trace
+enable_const_eval: False
--- a/blacksmith/experiments/torch/llama/xla/train.py
+++ b/blacksmith/experiments/torch/llama/xla/train.py
@@ -199,13 +199,14 @@ def train(
                     torch_xla._XLAC._xla_sync_multi(tensors_to_sync, devices, wait=True)
                     # Optimizer will do unnecessary recompute if we don't clear the pending IRs after syncing the loss and grads.
                     torch_xla._XLAC._clear_pending_irs(torch_xla._XLAC._xla_get_default_device())
+                
+                running_loss += loss_.item()
+                accumulation_step += 1
 
                 # Only step the optimizer after accumulating gradients.
                 if accumulation_step == config.gradient_accumulation_steps:
                     device_manager.optimizer_step(optimizer)
 
-                    running_loss += step_loss.item()
-                    step_loss = None
                     accumulation_step = 0
                     global_step += 1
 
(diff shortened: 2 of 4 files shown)


blacksmith/experiments/torch/llama/xla/lora/single_chip/llama_3_1_8b_sst2.yaml
```diff
@@ -3,13 +3,13 @@ dataset_id: "sst2"
...
 model_name: "meta-llama/Llama-3.1-8B"
-max_length: 128
+max_length: 512
 dtype: "torch.bfloat16"
 
 # Training hyperparameters
 training_model_type: "lora"
 learning_rate: 1e-4
-batch_size: 8
+batch_size: 1
```
pglusacTT: Not sure if we should change hyperparams from existing experiments?

agobeljicTT: does it fail with bs=8

agobeljicTT (commented):
Do we want to have this toggle for other models? Perhaps a good idea to do it all in this pr and just flip for llama

blacksmith/tools/templates/configs.py
```diff
@@ -78,3 +78,4 @@ class TrainingConfig(BaseModel):
     use_tt: bool = Field(default=True)
     enable_trace: bool = Field(default=False)
     trace_region_size: int = Field(default=1000000000, gt=0)  # DRAM region size (bytes) for runtime trace
+    enable_const_eval: bool = Field(default=True)
```
mmilosevicTT: This is something we should agree on, but I would like to decouple template config from experiment configs. The thing is that we already do not extend this in experiments and it becomes really hard to follow what is used for debugging and what for production. Also it is a bit weird to have templates where we have a config, but we do not update template train script to use it as well. @pglusacTT wdyt?

blacksmith/experiments/torch/llama/xla/lora/single_chip/llama_3_1_8b_sst2.yaml
abogdanovicTT: By default, bs=8, seq=128. These changes bs=1, seq=512 (could be update to 1024, once https://github.com/tenstorrent/tt-blacksmith/issues/625 is resolved) are only to keep it in sync with a perf measurements. What do you think, should we update it or keep the default values?

blacksmith/tools/templates/configs.py
abogdanovicTT: I agree. We have discussed this before as well. In the current setup, the experiment config uses BaseModel, and the template config is not being used. In any case, we should decide/refactor the experiments to use the template config in that case.

blacksmith/experiments/torch/llama/xla/lora/single_chip/llama_3_1_8b_sst2.yaml
agobeljicTT: personally i don't think that this is an issue, this just tilts our CI, but i'm ok with this
check if we need to recalc golden files and that CI is not broken by this

blacksmith/tools/templates/configs.py
agobeljicTT: good idea, but that does deserve a dedicated PR

blacksmith/tools/templates/configs.py
```diff
@@ -83,6 +83,7 @@ class TrainingConfig(BaseModel):
     use_tt: bool = Field(default=True)
     enable_trace: bool = Field(default=False)
     trace_region_size: int = Field(default=1000000000, gt=0)  # DRAM region size (bytes) for runtime trace
+    enable_const_eval: bool = Field(default=False)
```
agobeljicTT: True by default?

blacksmith/experiments/torch/llama/xla/train.py
```diff
@@ -166,6 +166,7 @@ def train(
         for epoch in range(config.num_epochs):
             accumulation_step = 0
             running_loss = 0.0
+            step_loss = None
```
agobeljicTT: remove

agobeljicTT (approved):

blacksmith/tools/templates/configs.py
abogdanovicTT: Sure, thanks!

PR #662, opened 2026-07-06
Add configurable tt-mlir optimization level to torch experiments

agobeljicTT:
### Issue
N/A

### Problem description
Torch training experiments had no way to set the tt-mlir optimizer level; it was implicitly 0 (all optimizations off), with wan2_2 hardcoding it.

### What's Changed
- Added an `optimization_level` config field (0/1/2, default 0) to all torch experiment configs.
- Wired for Llama, Qwen and Wan.
- Default 0 keeps existing behavior unchanged. Passed as a string to match tt-xla's convention. llama/ffe (forge backend) is out of scope.

### Checklist


blacksmith/experiments/torch/albert/configs.py
```diff
@@ -77,4 +77,5 @@ class TrainingConfig(BaseModel):
     # Other settings
     framework: str = Field(default="pytorch")
     use_tt: bool = Field(default=True)
+    optimization_level: int = Field(default=0, ge=0, le=2)
```
mmilosevicTT: You are missing this in `train.py`.

pglusacTT:
For now, can we add this for the perf target models in order to keep the diff minimal? We can include this optimization level knob repo-wdie once we start integrating the Trainer class we are working on.

blacksmith/experiments/torch/BOUNTIES/ppo_breakout/README.md
```diff
@@ -22,7 +22,7 @@ itself, along with the use of GAE for advantage estimation, is from Schulman et
 ## Training
 
 ```bash
-python blacksmith/experiments/torch/BOUNTIES/ppo_breakout/test_breakout_ppo_training.py
+python blacksmith/experiments/torch/BOUNTIES/ppo_breakout/train.py
```
pglusacTT, line 25: This doesn't feel like it should belong in this PR?

mmilosevicTT:
> For now, can we add this for the perf target models in order to keep the diff minimal? We can include this optimization level knob repo-wdie once we start integrating the Trainer class we are working on.

Agree with this, optimized/debuggable version of trainer class should be added as well, I have that on my TODO list, so you guys can test out features in a quicker way.

blacksmith/experiments/torch/BOUNTIES/ppo_breakout/README.md
agobeljicTT: it doesn't but it can stay, no need for an extra PR just for this file rename...

agobeljicTT:
> > For now, can we add this for the perf target models in order to keep the diff minimal? We can include this optimization level knob repo-wdie once we start integrating the Trainer class we are working on.
> 
> Agree with this, optimized/debuggable version of trainer class should be added as well, I have that on my TODO list, so you guys can test out features in a quicker way.

I added just the field everywhere, only a few experiments have actually wired it.

mmilosevicTT (approved):
LGTM

PR #641, opened 2026-06-25
[Trainer] Add LoraLLMTrainer and Default Training Structure

mmilosevicTT:
### Issue
Closes https://github.com/tenstorrent/tt-blacksmith/issues/617

### Problem description
This PR adds training through trainer class for lora llm models.

### What's Changed
- Added a generic `Trainer` base class (`blacksmith/tools/trainer/trainer.py`) that orchestrates the training loop: epochs, gradient accumulation, inline validation, optimizer setup, callback hooks, and device setup via `DeviceManager`.
- Integrated `ReproducibilityManager` into `Trainer.setup()` so seeds are set automatically from config (no manual seeding in scripts).
- Added `LoraLLMTrainer` strategy for parameter-efficient fine-tuning of causal LLMs (model/dataloader loading + forward/loss).
- Introduced a trainer config hierarchy: `TrainerConfig` (base) and `LoraLLMConfig`.
- Added an example runner (`examples/lora_llm/train.py`) plus SST-2 + Llama-3.2-1B configs for single-chip, tensor-parallel, and data-parallel setups.
- Updated the callback system to support the trainer lifecycle hooks.

### Checklist
- [x] New/Existing tests provide coverage for changes

PR #670, opened 2026-07-15
[Trainer] Add Late Feedback to Trainer

mmilosevicTT:
### Issue
N/A

### Problem description
There were several comments left by @ndrakulicTT on https://github.com/tenstorrent/tt-blacksmith/pull/641 post merge, so this PR addresses all the issues.

### What's Changed
This PR changes:
- Remove need for eval in dtype
- Fix use_cache config setting
- Add abstraction with contextmanager for try-catch-finally
- Remove defaults from .py config files
- Add option for different parameters in callbacks
- Make experiment configs extend template config

### Checklist
- [x] New/Existing tests provide coverage for changes

### DIFF
--- a/blacksmith/experiments/torch/wan2_2/configs.py
+++ b/blacksmith/experiments/torch/wan2_2/configs.py
@@ -6,9 +6,9 @@ from typing import List, Optional, Tuple
 import torch
 from pydantic import BaseModel, Field, model_validator
 
-from blacksmith.tools.templates.configs import Framework
+from blacksmith.tools.templates.configs import TrainingConfig as BaseTrainingConfig
 
-_TORCH_DTYPES = {
+_VAE_TORCH_DTYPES = {
     "torch.bfloat16": torch.bfloat16,
     "torch.float32": torch.float32,
     "torch.float16": torch.float16,
@@ -45,16 +45,15 @@ class InferenceConfig(BaseModel):
         return self
 
 
-class TrainingConfig(BaseModel):
+class TrainingConfig(BaseTrainingConfig):
     # Selects what train.py runs: "train" fine-tunes the LoRA on cached latents;
     # "infer" loads a checkpoint and generates a video (no training).
     mode: str = Field(default="train")
 
     # Model settings
     model_id: str = Field(default="Wan-AI/Wan2.2-TI2V-5B-Diffusers")
-    dtype: str = Field(default="torch.bfloat16")
+    model_name: str = Field(default="Wan-AI/Wan2.2-TI2V-5B-Diffusers")
     vae_precompute_dtype: str = Field(default="torch.bfloat16")
-    gradient_checkpointing: bool = Field(default=False)
 
     # Dataset / cache settings
     dataset_id: str = Field(default="jainr3/diffusiondb-pixelart")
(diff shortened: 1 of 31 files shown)


blacksmith/experiments/torch/wan2_2/configs.py
ndrakulicTT, line 19: Just noticed this. For the future question:
Do we have plan to offload this to inference server for example?

blacksmith/tools/trainer/examples/lora_llm/data_parallel/llama_3_2_1b_sst2.yaml
```diff
@@ -7,11 +7,13 @@ max_length: 64
 dtype: "torch.bfloat16"
 
 # Training hyperparameters
+training_model_type: "lora"
 learning_rate: 1e-4
 batch_size: 16
 gradient_accumulation_steps: 1
 num_epochs: 1
-ignored_index: -100
```
ndrakulicTT, line 14: Why did we remove this config?

blacksmith/tools/trainer/configs/base.py
```diff
@@ -16,38 +22,43 @@ class TrainerConfig(BaseModel):
...
+    deterministic: bool
 
     # Device / sharding settings
     use_tt: bool = Field(default=True)
     mesh_shape: Optional[list[int]] = Field(default=None)
     mesh_axis_names: Optional[list[str]] = Field(default=None)
     input_sharding_dim: Optional[str] = Field(default=None)
     model_sharding_patterns: Optional[List[Tuple[str, Tuple[Optional[str], ...]]]] = Field(default=None)
+
+    def torch_dtype(self) -> torch.dtype:
```
ndrakulicTT, line 60: Nice, I like the solution, if we ever need for the jax filed is still dtype in config you can just add jax_dtype.
And more importantly it is simple

blacksmith/tools/templates/configs.py
ndrakulicTT, line 1: Do we still need this template folder?

blacksmith/tools/trainer/trainer.py
```diff
@@ -86,14 +87,27 @@ def _load_optimizer(self) -> torch.optim.Optimizer:
             capturable=self.config.use_tt,
         )
 
+    @contextmanager
+    def _train_lifecycle(self) -> Iterator[None]:
+        """Run error / cleanup hooks around the training loop body."""
+        try:
+            yield
+        except Exception as exception:
+            self.callback_handler("on_error", exception)
```
ndrakulicTT, line 96: Do we want to swallow exception here?

ndrakulicTT (approved):

blacksmith/tools/trainer/callback.py
```diff
@@ -6,56 +6,56 @@
...
-    def on_validation_batch_start(self, trainer, batch):
+    def on_validation_batch_start(self, trainer, batch, *args, **kwargs):
         pass
 
-    def on_validation_batch_end(self, trainer, batch, loss):
+    def on_validation_batch_end(self, trainer, batch, loss, *args, **kwargs):
         pass
 
-    def on_validation_end(self, trainer, val_loss):
+    def on_validation_end(self, trainer, *args, **kwargs):
```
ndrakulicTT: Do we want here: `val_loss, *args, **kwargs`? to match the rest of methods

blacksmith/tools/trainer/trainer.py
mmilosevicTT: I do not see why not? I mean you might want to add some additional processing of error on the fly.

blacksmith/tools/templates/configs.py
mmilosevicTT: Yes, how I look at this is the following. We have a template config, you can use it and then extend it to your needs to do what the dir says - EXPERIMENT. But then we keep structured format in another config for trainer which has to be more concise.

blacksmith/tools/trainer/examples/lora_llm/data_parallel/llama_3_2_1b_sst2.yaml
mmilosevicTT: LoRALLM so far always used ignore_index = -100, we have not seen any change it this.

blacksmith/experiments/torch/wan2_2/configs.py
mmilosevicTT: I think they already have a good loop to do image generation. This here is just for the purpose of testing if training is working. @umalesTT am I correct?

Also, I would say this one has to have separate trainer class, as it is videoGen model anyways

blacksmith/tools/trainer/callback.py
mmilosevicTT: We can keep it for now actually, and if it turns out we do not need it in general, we can always just clean it up quickly. Will revert to val_loss as a separate arg.

blacksmith/experiments/torch/wan2_2/configs.py
umalesTT: Yes, exactly @mmilosevicTT that is the purpose
+1 on separate trainer class

PR #673, opened 2026-07-16
[PERF] Fuse forward + backward + optimizer into a single graph

abogdanovicTT:
### Issue

N/A

### Problem description

The training step runs the forward+backward and the optimizer as two separate graphs. Two things force extra compilations / executions:

1. After `loss_.backward()` the loss and all grads are synced with `_xla_sync_multi` + `_clear_pending_irs`, which cuts the graph between backward and the optimizer step. Forward+backward and the optimizer update therefore compile and execute as separate graphs every step.
2. `torch.optim.AdamW` lazily allocates its `step` / `exp_avg` / `exp_avg_sq` state on the first `step()`. This changes the optimizer graph's input signature after step 0 and forces a second compilation of the step.

This PR fuses forward + backward + optimizer into a single graph and pre-materializes the AdamW state so the fused step compiles only once.

### What's Changed

- Fuse the whole step into one graph in `train.py`: drop the intermediate `_xla_sync_multi` / `_clear_pending_irs` on loss+grads so the forward/backward IR stays pending and joins the optimizer update, executed by the single sync inside `optimizer_step`.
- Add `materialize_adamw_state` to `tools/workaround_utils.py`: pre-initializes AdamW `step`/`exp_avg`/`exp_avg_sq` to zero (a numerical no-op, since Adam's own lazy init also starts at zero) so the fused step graph keeps a stable input signature and compiles once.
- Accumulate a detached `step_loss` lazily and read it (`.item()`) only after `optimizer_step`, so logging the loss no longer cuts the graph.

### DIFF
--- a/blacksmith/experiments/torch/llama/xla/lora/single_chip/llama_3_1_8b_sst2.yaml
+++ b/blacksmith/experiments/torch/llama/xla/lora/single_chip/llama_3_1_8b_sst2.yaml
@@ -3,13 +3,13 @@ dataset_id: "sst2"
 
 # Model settings
 model_name: "meta-llama/Llama-3.1-8B"
-max_length: 512
+max_length: 128
 dtype: "torch.bfloat16"
 
 # Training hyperparameters
 training_model_type: "lora"
 learning_rate: 1e-4
-batch_size: 1
+batch_size: 8
 num_epochs: 1
 ignored_index: -100
 
--- a/blacksmith/experiments/torch/llama/xla/lora/single_chip/llama_3_1_8b_sst2.yaml
+++ b/blacksmith/experiments/torch/llama/xla/lora/single_chip/llama_3_1_8b_sst2.yaml
@@ -3,13 +3,13 @@ dataset_id: "sst2"
 
 # Model settings
 model_name: "meta-llama/Llama-3.1-8B"
-max_length: 128
+max_length: 512
 dtype: "torch.bfloat16"
 
 # Training hyperparameters
 training_model_type: "lora"
 learning_rate: 1e-4
-batch_size: 8
+batch_size: 1
 num_epochs: 1
 ignored_index: -100
 
@@ -59,3 +59,4 @@ use_tt: True
 measure_e2e_time: False
 enable_trace: False
 trace_region_size: 1000000000 # DRAM region size (bytes) for runtime trace
+enable_const_eval: False
--- a/blacksmith/experiments/torch/llama/xla/lora/single_chip/llama_3_1_8b_sst2.yaml
+++ b/blacksmith/experiments/torch/llama/xla/lora/single_chip/llama_3_1_8b_sst2.yaml
@@ -3,13 +3,13 @@ dataset_id: "sst2"
 
 # Model settings
 model_name: "meta-llama/Llama-3.1-8B"
-max_length: 512
+max_length: 128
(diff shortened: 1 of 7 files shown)


blacksmith/experiments/torch/llama/xla/lora/single_chip/llama_3_1_8b_sst2.yaml
```diff
@@ -32,8 +32,8 @@ wandb_tags: ["test"]
 wandb_watch_mode: "all"
 wandb_log_freq: 1000
 model_to_wandb: False
-steps_freq: 10
-val_steps_freq: 10
```
mmilosevicTT, line 36: This does not seem related to the PR.

blacksmith/experiments/torch/llama/xla/train.py
```diff
@@ -192,20 +201,20 @@ def train(
...
-                    # Optimizer will do unnecessary recompute if we don't clear the pending IRs after syncing the loss and grads.
-                    torch_xla._XLAC._clear_pending_irs(torch_xla._XLAC._xla_get_default_device())
-
-                running_loss += loss_.item()
+                # Keep the whole step (forward + backward + optimizer) in a single graph.
+                # No sync is issued here on purpose: the forward and backward IR stays pending
+                # and lazily joins the optimizer update, so everything is compiled and executed
+                # together by the one sync inside `optimizer_step`. The detached micro-batch loss
+                # is accumulated lazily so it can be read afterwards without cutting the graph.
+                step_loss = loss_.detach() if step_loss is None else step_loss + loss_.detach()
```
mmilosevicTT: So this means we are extending the graph to grad_accumulation_step loops + optimizer? Will this pose problems with memory?

blacksmith/experiments/torch/llama/xla/train.py
```diff
@@ -113,6 +117,10 @@ def train(
     trainable_params = [p for p in model.parameters() if p.requires_grad]
     optimizer = torch.optim.AdamW(trainable_params, capturable=config.use_tt, lr=config.learning_rate)
 
+    # Pre-create optimizer state so the fused step graph compiles only once.
+    if config.use_tt and not config.resume_from_checkpoint:
+        materialize_adamw_state(optimizer)
+
     # Load checkpoint if needed.
     if config.resume_from_checkpoint:
         checkpoint_manager.load_checkpoint(model, optimizer)
```
mmilosevicTT, line 141: Did you try loading from checkpoint? Will this always stay one graph? i.e does load_checkpoint not introduce some operations like it was with fresh optimizer?

mmilosevicTT (approved):
LGTM

Left few comments

agobeljicTT (approved):

blacksmith/experiments/torch/llama/xla/train.py
agobeljicTT: fixed it in the last commit, now we have one fwd+bwd and one fwd+bwd+opt

blacksmith/experiments/torch/llama/xla/train.py
agobeljicTT: i'll try this Monday, make myself a reminder

agobeljicTT: there was an issue, resolved it by putting optimizer to device and syncing

PR #683, opened 2026-07-28
[Trainer] Add Option for Custom Dataset in Trainer Class

### LINKED ISSUE #619: Add Option for Custom Dataset in Trainer Class
### Is your feature request related to a problem? Please describe

Although most of the training through TT-Blacksmith is through config specific experiments, we want to enable trainer class to pass in custom datasets that will be prepared by customers.

### Describe the solution you'd like

_No response_

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

tstepanovicTT:
## Issue
Closes https://github.com/tenstorrent/tt-blacksmith/issues/619

## Problem description

The trainer currently supports predefined datasets only, so user-provided local datasets cannot be plugged into the standard training flow. This change adds support for custom JSON/JSONL datasets, with separate train and optional validation files, so custom data can be loaded and used through the existing trainer pipeline instead of requiring a new dataset-specific implementation.

## What's changed

- Added a new `custom` dataset option in dataset selection, so `get_dataset(config, split=...)` routes both train and validation requests through the standard trainer entry point.
- Added `CustomDatasetConfig` to define file type, train/validation file paths, prompt format, and column mapping for local custom datasets.
- Integrated custom datasets into the trainer data flow: the selected split is loaded from local files with `load_dataset`, the configured columns are validated, each example is converted into a prompt/response pair in the given format (e.g. Alpaca-style), tokenized, filtered by `max_length`, reduced to trainer-required columns, and exposed through the standard dataloader path. The train split is shuffled, and if no validation file is provided, the validation dataset/dataloader is skipped.
- Added small utilities for file type normalization and column-mapping validation to make custom file loading more robust. 

## Training curves

Loss curves below are from a trainer run on a local split of Databricks Dolly 15k using the custom dataset path.

### Train loss
pr_train

### Validation loss
pr_val

### Train loss when there is no validation set
pr_train_no_val

### DIFF
--- a/blacksmith/datasets/torch/custom/custom_dataset_utils.py
+++ b/blacksmith/datasets/torch/custom/custom_dataset_utils.py
@@ -2,6 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 from enum import Enum
+from typing import Dict
 
 from blacksmith.datasets.torch.custom.templates.alpaca import build_alpaca_prompt
 
@@ -10,10 +11,50 @@ class AvailableFormats(Enum):
     ALPACA = "alpaca"
 
 
-def build_prompt(format: str, instruction: str, input_text: str = ""):
+FORMAT_REQUIRED_KEYS = {
+    AvailableFormats.ALPACA.value: {"instruction", "output"},
+}
+
+
+def build_prompt(format: str, column_mapping: Dict, example: Dict):
     format = format.strip().lower()
+
     if format == AvailableFormats.ALPACA.value:
-        return build_alpaca_prompt(instruction, input_text)
+        return build_alpaca_prompt(column_mapping, example)
     else:
         available_formats = [f.value for f in AvailableFormats]
-        raise ValueError(f"Unsupported format: {format}. Available formats: {available_formats}")
+        raise ValueError(
+            f"Selected format is unsupported: {format}. You should use one of the available formats: {available_formats}"
+        )
+
+
+def normalize_file_type(file_type: str) -> str:
+    file_type = file_type.strip().lower()
+    mapping = {
+        "json": "json",
+        "jsonl": "json",
+    }
+    if file_type not in mapping:
+        raise ValueError(
+            f"Selected file type is unsupported: {file_type}. "
+            f"Please select one of the supported types: {list(mapping)}"
+        )
+    return mapping[file_type]
+
+
+def validate_column_mapping(format: str, column_mapping: Dict, dataset_columns: set):
+    # 1. Required keys must be present and non-empty
+    required_keys = FORMAT_REQUIRED_KEYS[format]
+    missing_keys = required_keys - column_mapping.keys()
+    if missing_keys:
+        raise ValueError(
+            f"column_mapping is missing required key(s): {sorted(missing_keys)}. " f"Required: {sorted(required_keys)}."
+        )
+    empty_required = [k for k in required_keys if not column_mapping.get(k)]
+    if empty_required:
+        raise ValueError(f"Required key(s) have empty mapping: {sorted(empty_required)}.")
+
+    # 2. Every non-empty mapped value must exist as an actual dataset column
+    for col in column_mapping.values():
+        if col and col not in dataset_columns:
+            raise ValueError(f"Column '{col}' not found in dataset columns: {sorted(dataset_columns)}.")
--- a/blacksmith/datasets/torch/custom/custom_dataset_utils.py
+++ b/blacksmith/datasets/torch/custom/custom_dataset_utils.py
@@ -1,0 +1,33 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+from string import Template
+
+ALPACA_PROMPT_INTRO = (
+    "Below is an instruction that describes a task. Write a response that appropriately completes the request."
+)
+ALPACA_PROMPT_TEMPLATE = Template(
+    f"""
+{ALPACA_PROMPT_INTRO}
+
+### Instruction:
+$instruction
+
+$input_section
+
+### Response:
+"""
+)
+ALPACA_INPUT_BLOCK_TEMPLATE = Template(
+    """### Input:
+$input
+"""
+)
+
+
+def build_alpaca_prompt(instruction: str, input_text: str = "") -> str:
+    input_section = ALPACA_INPUT_BLOCK_TEMPLATE.substitute(input=input_text) if input_text.strip() else ""
+    return ALPACA_PROMPT_TEMPLATE.substitute(
+        instruction=instruction,
+        input_section=input_section,
+    )
--- a/blacksmith/datasets/torch/custom/custom_dataset_utils.py
+++ b/blacksmith/datasets/torch/custom/custom_dataset_utils.py
@@ -1,28 +1,9 @@
 # SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
 #
 # SPDX-License-Identifier: Apache-2.0
-from string import Template
-
-ALPACA_PROMPT_INTRO = (
-    "Below is an instruction that describes a task, paired with an optional input that provides further context. "
-    "Write a response that appropriately completes the request."
-)
-ALPACA_PROMPT_TEMPLATE = Template(
-    f"""
-{ALPACA_PROMPT_INTRO}
-
-### Instruction:
-$instruction
-
-$input_section
-
-### Response:
-"""
-)
-ALPACA_INPUT_BLOCK_TEMPLATE = Template(
-    """### Input:
-$input
-"""
+from blacksmith.datasets.torch.custom.templates.alpaca import (
+    ALPACA_INPUT_BLOCK_TEMPLATE,
+    ALPACA_PROMPT_TEMPLATE,
 )
 
 
--- a/blacksmith/datasets/torch/custom/custom_dataset_utils.py
+++ b/blacksmith/datasets/torch/custom/custom_dataset_utils.py
@@ -1,15 +1,19 @@
 # SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
 #
 # SPDX-License-Identifier: Apache-2.0
-from blacksmith.datasets.torch.custom.templates.alpaca import (
-    ALPACA_INPUT_BLOCK_TEMPLATE,
-    ALPACA_PROMPT_TEMPLATE,
-)
+from enum import Enum
+
+from blacksmith.datasets.torch.custom.templates.alpaca import build_alpaca_prompt
 
 
-def build_alpaca_prompt(instruction: str, input_text: str = "") -> str:
-    input_section = ALPACA_INPUT_BLOCK_TEMPLATE.substitute(input=input_text) if input_text.strip() else ""
-    return ALPACA_PROMPT_TEMPLATE.substitute(
-        instruction=instruction,
-        input_section=input_section,
-    )
+class AvailableFormats(Enum):
+    ALPACA = "alpaca"
+
+
+def build_prompt(format: str, instruction: str, input_text: str = ""):
+    format = format.strip().lower()
+    if format == AvailableFormats.ALPACA.value:
+        return build_alpaca_prompt(instruction, input_text)
+    else:
+        available_formats = [f.value for f in AvailableFormats]
+        raise ValueError(f"Unsupported format: {format}. Available formats: {available_formats}")
--- a/blacksmith/datasets/torch/custom/custom_dataset_utils.py
+++ b/blacksmith/datasets/torch/custom/custom_dataset_utils.py
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 from enum import Enum
-from typing import Dict
+from typing import Dict, Optional
 
 from blacksmith.datasets.torch.custom.templates.alpaca import build_alpaca_prompt
 
@@ -11,16 +11,14 @@ class AvailableFormats(Enum):
     ALPACA = "alpaca"
 
 
-FORMAT_REQUIRED_KEYS = {
-    AvailableFormats.ALPACA.value: {"instruction", "output"},
-}
+FORMAT_KEYS = {AvailableFormats.ALPACA.value: {"required": {"instruction", "output"}, "optional": {"input"}}}
 
 
-def build_prompt(format: str, column_mapping: Dict, example: Dict):
+def build_prompt(example: Dict, format: str, column_mapping: Dict) -> tuple[str, str, str]:
     format = format.strip().lower()
 
     if format == AvailableFormats.ALPACA.value:
-        return build_alpaca_prompt(column_mapping, example)
+        return build_alpaca_prompt(example, column_mapping)
     else:
         available_formats = [f.value for f in AvailableFormats]
         raise ValueError(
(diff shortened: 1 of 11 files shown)


blacksmith/tools/configs.py
```diff
@@ -73,3 +75,26 @@ class CheckpointConfig(BaseModel):
...
+    column_mapping: Optional[Dict[str, str]] = Field(
+        default_factory=lambda: {
+            "instruction": "instruction",
+            "input": "input",
+            "output": "output",
+        }
+    )
+
+    # Whether to drop the last batch
+    drop_last: bool = Field(default=True)
```
mmilosevicTT: I would for now force dropping of last batch, would not allow keeping it.

blacksmith/tools/templates/configs.py
```diff
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 from enum import Enum
-from typing import Optional
+from typing import Literal, Optional
```
mmilosevicTT: nit: just remove this unnecessary change

blacksmith/datasets/torch/custom/custom_dataset_utils.py
```diff
@@ -0,0 +1,34 @@
...
+$input_section
+
+### Response:
+"""
+)
+ALPACA_INPUT_BLOCK_TEMPLATE = Template(
+    """### Input:
+$input
+"""
+)
```
mmilosevicTT: These might be better to be in the separate dir file like: `blacksmith/datasets/torch/custom/template/alpaca.py`. Then when can have a helper function that will call the builder of prompt based on input config.

mmilosevicTT (commented):
All looks good so far. Left few comments.

blacksmith/datasets/torch/custom/custom_llm_dataset.py
```diff
@@ -0,0 +1,94 @@
...
+        super().__init__(config, split, collate_fn)
+
+    def _format_example(self, example):
+        # TODO(tstepanovicTT): Make it work for any prompt format, not just Alpaca.
+        instruction = example[self.column_mapping["instruction"]]
+        input_col = self.column_mapping.get("input", "")
+        input_text = example[input_col] if input_col else ""
+        output = example[self.column_mapping["output"]]
+
+        prompt = build_alpaca_prompt(instruction, input_text)
```
mcupacTT: Here, we want build_prompt only, right?

tstepanovicTT: Yes, I have fixed that in the meantime

blacksmith/tools/configs.py
```diff
@@ -73,3 +75,29 @@ class CheckpointConfig(BaseModel):
...
+
+
+class CustomDatasetConfig(BaseModel):
+    """
+    Additional config in case of custom datasets.
+    Train and validation sets should be loaded from separate files.
+    The validation set is optional.
+    """
+
+    file_type: Literal["json", "jsonl"] = Field(default="json")
```
mcupacTT: it seems that HF datasets has no jsonl loader - but you can double check

blacksmith/datasets/torch/custom/custom_llm_dataset.py
```diff
@@ -0,0 +1,102 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+from torch.utils.data import DataLoader
+from transformers import AutoTokenizer, DataCollatorForSeq2Seq
+
+from blacksmith.datasets.torch.custom.custom_dataset_utils import build_prompt
+from blacksmith.datasets.torch.torch_dataset import BaseDataset
+from blacksmith.tools.templates.configs import TrainingConfig
```
mcupacTT: here we should import TrainerConfig and not TrainingConfig?

blacksmith/tools/trainer/configs/base.py
```diff
@@ -40,10 +45,11 @@ class TrainerConfig(BaseModel):
     # Validation settings
     val_steps_freq: int = Field(gt=0)
 
-    # Logging / metrics / checkpointing settings (nested sub-configs).
+    # Logging / metrics / checkpointing / custom dataset settings (nested sub-configs).
     logging: LoggingConfig
     metrics: MetricsConfig
     checkpoint: CheckpointConfig
+    dataset: CustomDatasetConfig
```
mcupacTT: this should be optional? In case we do not use CustomDatasetConfig

also, add pydantic validator that this field is enforced if dataset_id is custom

blacksmith/datasets/torch/custom/custom_llm_dataset.py
```diff
@@ -0,0 +1,102 @@
...
+        self.column_mapping = config.dataset.column_mapping
+
+        super().__init__(config, split, collate_fn)
+
+    def _format_example(self, example):
+        # TODO(tstepanovicTT): Make it work for any prompt format, not just Alpaca.
+        instruction = example[self.column_mapping["instruction"]]
+        input_col = self.column_mapping.get("input", "")
+        input_text = example[input_col] if input_col else ""
+        output = example[self.column_mapping["output"]]
```
mcupacTT: we should add validadation here to raise error message in case something is not given as expected:
1. in case mapped column doesnt exist
2. or the required key like output is missing from the config
3. or column_mapping is None

mcupacTT (commented):
Left few comments!

blacksmith/datasets/torch/custom/templates/alpaca.py
```diff
@@ -0,0 +1,34 @@
...
+$input_section
+
+### Response:
+"""
+)
+ALPACA_INPUT_BLOCK_TEMPLATE = Template(
+    """### Input:
+$input
+"""
+)
```
mmilosevicTT: nit: I would prefer Input block being before prompt template as you need former to build latter.

blacksmith/datasets/torch/custom/custom_dataset_utils.py
```diff
@@ -0,0 +1,19 @@
...
+    ALPACA = "alpaca"
+
+
+def build_prompt(format: str, instruction: str, input_text: str = ""):
+    format = format.strip().lower()
+    if format == AvailableFormats.ALPACA.value:
+        return build_alpaca_prompt(instruction, input_text)
+    else:
+        available_formats = [f.value for f in AvailableFormats]
+        raise ValueError(f"Unsupported format: {format}. Available formats: {available_formats}")
```
mmilosevicTT: nit: Let's make error message more user friendly:
```suggestion
        raise ValueError(f"Selected format is unsupported: {format}. You should use one of the available formats: {available_formats}")
```

blacksmith/datasets/torch/custom/custom_llm_dataset.py
mmilosevicTT: +1 on this, I would move away from this being used in experiments. Python does not do proper inheritance enforcing, but I would not rely on TrainingConfig which has many different things.

blacksmith/datasets/torch/custom/custom_llm_dataset.py
mmilosevicTT: First two I agree, last one I would just expect the default from template. I would also add that this should probably go to build_prompt function. For LLMs I would always expect to have prompt+output+full_text as outputs of preparing.

blacksmith/datasets/torch/custom/custom_llm_dataset.py
```diff
@@ -0,0 +1,102 @@
...
+        example["attention_mask"] = attention_mask
+        example["labels"] = labels
+        example["len"] = input_ids.size(0)
+
+        return example
+
+    def _prepare_dataset(self):
+        if self.dataset_path is not None:
+            # split="train" is the HF split label for the loaded file, does not mean the file contains training data
+            raw_dataset = load_dataset(self.file_type, data_files=self.dataset_path, split="train")
```
mmilosevicTT: do we need the split part?

blacksmith/tools/trainer/configs/base.py
mmilosevicTT: +1 for the comment

blacksmith/tools/trainer/examples/lora_llm/custom_datasets/llama_3_2_1b_custom.yaml
```diff
@@ -0,0 +1,82 @@
+# Dataset settings
+dataset_id: "custom"
+dataset:
+  file_type: "json"
+  train_dataset_path: "/path/to/local/train.json"
+  val_dataset_path: "/path/to/local/val.json"
+  format: "alpaca"
+  column_mapping: { # change accordingly
```
mmilosevicTT: nit: This comment is not really useful here.
```suggestion
  column_mapping: {
```

blacksmith/tools/trainer/examples/lora_llm/single_chip/llama_3_2_1b_custom.yaml
mmilosevicTT, line 1: I would not add dir `custom_datasets` here. Configs are usually changed based on underlying chip throughout the repo.

blacksmith/tools/configs.py
mmilosevicTT: By this I mean, I would remove this field from the config for now.

blacksmith/datasets/torch/custom/custom_llm_dataset.py
```diff
@@ -0,0 +1,102 @@
...
+            filtered_dataset = tokenized_dataset.filter(lambda x: x["len"] <= self.config.max_length)
+            filtered_dataset = filtered_dataset.remove_columns(
+                [col for col in filtered_dataset.column_names if col not in self.required_columns]
+            )
+            if self.split == "train":
+                filtered_dataset = filtered_dataset.shuffle(seed=self.config.seed)
+
+            self.dataset = filtered_dataset
+        else:
+            self.dataset = None
```
mmilosevicTT, line 66: I would error out if we enter here with train path, as that means we do not have a dataset. Also please do early return. something like:
```
if self.dataset_path is None:
    if self.split == "train":
        error-out
    self.dataset = None
    return
rest-of-the-code
```

blacksmith/datasets/torch/custom/custom_llm_dataset.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def __len__(self):
+        return len(self.dataset)
+
+    def _get_dataloader(self):
+        if self.dataset is not None:
```
mmilosevicTT: again early return:
```
if self.dataset is None:
    return None
rest-of-the-code
```

mmilosevicTT (commented):
Great work @tstepanovicTT. 

I second all comments from @mcupacTT and have left few more comments.

mmilosevicTT:
Could you please attach loss curve for train/val splits of the training you did through trainer, so we can have confirmation it works? Also, please attach closing of the https://github.com/tenstorrent/tt-blacksmith/issues/619 in the PR description and rename the title of the PR to match the issue title.

mmilosevicTT:
Additionally it would be useful to add a bit of description on how custom dataset is used in training scripts, not just what classes/objects we added.

blacksmith/datasets/torch/custom/custom_llm_dataset.py
tstepanovicTT: Actually yes, without it an error is raised. load_dataset creates a DatasetDict which uses "train" as the key by default, so if you try loading a dataset with no keys specified, it will be saved under the "train" key. The comment above is referring to that, I should rephrase it to make it clearer.

blacksmith/tools/configs.py
tstepanovicTT: That's right, but we still want to be able to load both JSON and JSONL files. load_datasets() can load both of them using file_type="json" so I added a function normalize_file_type() to make sure JSONL file reading runs smoothly.

env/xla_requirements.txt
```diff
@@ -1,5 +1,5 @@
 --extra-index-url https://pypi.eng.aws.tenstorrent.com
 --extra-index-url https://download.pytorch.org/whl/cpu
-pjrt-plugin-tt==1.4.0.dev20260730001900
```
mmilosevicTT: This should not be done in this PR, no?

blacksmith/tools/trainer/configs/base.py
```diff
@@ -45,6 +50,17 @@ class TrainerConfig(BaseModel):
...
 
+    # Custom dataset settings.
+    custom_dataset: Optional[CustomDatasetConfig] = Field(default=None)
+
+    @model_validator(mode="after")
+    def check_custom_dataset(self) -> "TrainerConfig":
+        if self.dataset_id == "custom" and self.custom_dataset is None:
+            raise ValueError("`custom_dataset` is required when dataset_id='custom'")
+        if self.dataset_id != "custom" and self.custom_dataset is not None:
+            raise ValueError("`custom_dataset` is not available when dataset_id!='custom'")
```
mmilosevicTT: I would say this one is not required. We can have additional configs, just not use them.  In other case we need it, as we cannot build the dataset without configs.

blacksmith/tools/configs.py
```diff
@@ -73,3 +75,36 @@ class CheckpointConfig(BaseModel):
...
+    data_files: Dict[str, str] = Field(default_factory=dict)
+
+    @model_validator(mode="after")
+    def build_data_files(self) -> "CustomDatasetConfig":
+        if self.train_dataset_path is None or self.train_dataset_path == "":
+            raise ValueError("train_dataset_path is required and was not provided.")
+        self.data_files = {"train": self.train_dataset_path}
+        if self.val_dataset_path:
+            self.data_files["validation"] = self.val_dataset_path
+        return self
```
mmilosevicTT: I am not a fan of data_files being a part of the config. This is only needed if we do local file load with load_dataset, what happens if we introduce load from url from hf?

mmilosevicTT: I would prefer this being part of dataset preparation.

blacksmith/datasets/torch/custom/custom_llm_dataset.py
```diff
@@ -0,0 +1,100 @@
...
+
+        self.file_type = normalize_file_type(config.custom_dataset.file_type)
+        self.data_files = config.custom_dataset.data_files
+        self.format = config.custom_dataset.format
+        self.column_mapping = config.custom_dataset.column_mapping
+
+        super().__init__(config, split, collate_fn)
+
+    def _format_example(self, example):
+        return build_prompt(self.format, self.column_mapping, example)
```
mmilosevicTT: not really needed, no?

mmilosevicTT: We can just call it directly with build_prompt in the code.

blacksmith/datasets/torch/custom/custom_dataset_utils.py
```diff
@@ -0,0 +1,60 @@
...
+    mapping = {
+        "json": "json",
+        "jsonl": "json",
+    }
+    if file_type not in mapping:
+        raise ValueError(
+            f"Selected file type is unsupported: {file_type}. "
+            f"Please select one of the supported types: {list(mapping)}"
+        )
+    return mapping[file_type]
```
mmilosevicTT: I believe for now jsonl would be the only one that we would load with different input then what we get in config, could we do something like:
```suggestion
    if file_type == "jsonl":
        return "json"
    return file_type
```

If we come up to some other formats in future we can extend it.

blacksmith/datasets/torch/custom/custom_dataset_utils.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def validate_column_mapping(format: str, column_mapping: Dict, dataset_columns: set):
```
mmilosevicTT: This one I do not completely agree. I would allow column_mapping to be optional then we do following:
- If None -> just check that required keys are present in dataset columns
- If not None -> Do following:
1. Check if column_mapping values are present in dataset, if not fail out.
2. Check if column_mapping keys are a subset of all possible keys for the format, not just required.
3. If some required key was missing just do identity, if optional field was missing in column mapping, check if it is present in dataset columns, if yes do identity.
Wdyt?

blacksmith/datasets/torch/custom/templates/alpaca.py
```diff
@@ -0,0 +1,43 @@
...
+$instruction
+
+$input_section
+
+### Response:
+"""
+)
+
+
+def build_alpaca_prompt(column_mapping: Dict, example: Dict) -> str:
```
mmilosevicTT: Example first, I would make column_mapping optional.

mmilosevicTT (commented):
Left few comments, this is close to being done 😄 Great work.

blacksmith/datasets/torch/custom/custom_dataset_utils.py
tstepanovicTT: What happens with optional keys in case column_mapping is None? Is there a reason why we wouldn't do the same check as for the required keys?

env/xla_requirements.txt
```diff
@@ -1,5 +1,4 @@
 --extra-index-url https://pypi.eng.aws.tenstorrent.com
 --extra-index-url https://download.pytorch.org/whl/cpu
-pjrt-plugin-tt==1.5.0.dev20260804002041
```
mmilosevicTT, line 3: probably just missed, but do not remove this 😄

blacksmith/datasets/torch/custom/custom_dataset_utils.py
```diff
@@ -0,0 +1,86 @@
...
+from typing import Dict, Optional
+
+from blacksmith.datasets.torch.custom.templates.alpaca import build_alpaca_prompt
+
+
+class AvailableFormats(Enum):
+    ALPACA = "alpaca"
+
+
+FORMAT_KEYS = {AvailableFormats.ALPACA.value: {"required": {"instruction", "output"}, "optional": {"input"}}}
```
mmilosevicTT: can you reformat in a way:
```
{
    key1: {
                 key2: {
                              value1,
                              value2
                            },
                 key3: {
                              value3
                            }
              }
}
```

In this way, new entries would only mix with at most one new line.

blacksmith/datasets/torch/custom/custom_dataset_utils.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def normalize_file_type(file_type: str) -> str:
+    if file_type == "jsonl":
+        file_type = "json"
+    return file_type
+
+
+def resolve_column_mapping(
+    format_name: str, column_mapping: Optional[Dict[str, str]], dataset_columns: set[str]
```
mmilosevicTT: `format_name` -> `template_name`?

blacksmith/datasets/torch/custom/custom_dataset_utils.py
```diff
@@ -0,0 +1,86 @@
...
+    if format_name not in FORMAT_KEYS:
+        available_formats = [f.value for f in AvailableFormats]
+        raise ValueError(
+            f"Selected format is unsupported: {format_name}. "
+            f"You should use one of the available formats: {available_formats}"
+        )
+
+    required_keys = FORMAT_KEYS[format_name]["required"]
+    optional_keys = FORMAT_KEYS[format_name]["optional"]
+    all_possible_keys = required_keys.union(optional_keys)
```
mmilosevicTT: possible is not necessary, all keys is descriptive enough

blacksmith/datasets/torch/custom/custom_dataset_utils.py
```diff
@@ -0,0 +1,86 @@
...
+                f"Column mapping contains unsupported keys: {sorted(extra_keys)}. "
+                f"Supported keys for format '{format_name}': {sorted(all_possible_keys)}."
+            )
+
+    # Fill required keys by identity if possible
+    for key in required_keys:
+        if key in dataset_columns and key not in resolved:
+            resolved[key] = key
+
+    # Fill optional keys by identity if present
```
mmilosevicTT: nit:
```suggestion
    # Fill optional keys by identity if present in dataset columns
```

blacksmith/datasets/torch/custom/custom_dataset_utils.py
mmilosevicTT: Please refactor this so that indentation is a single tab.

mmilosevicTT (commented):
LGTM, please address final comments and we can merge this 😄

mmilosevicTT (approved):
LGTM 🎉 

Great work!

PR #704, opened 2026-08-08
Refactor Reproducibility Manager Not to Depend on Jax

mmilosevicTT:
### Issue
N/A

### Bug Description
Currently reproducibility manager is shared between jax and torch. This dependency makes it impossible to load tt-blacksmith's trainer only for torch purposes, without importing jax.

### What's Changed
Split reproducibility manager into two separate ones. In future if we decide to merge efforts, we can add base class that both extend.

### DIFF
--- a/blacksmith/tools/reproducibility_manager.py
+++ b/blacksmith/tools/reproducibility_manager.py
@@ -15,7 +15,7 @@ class ReproducibilityManager:
         self.config = config
 
     def setup(self):
-        self._setup_python()
+        self._seed_python_rngs()
 
         torch.manual_seed(self.config.seed)
         torch.cuda.manual_seed(self.config.seed)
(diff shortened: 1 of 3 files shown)


blacksmith/tools/reproducibility_manager.py
```diff
@@ -41,9 +25,6 @@ def _setup_pytorch(self):
             torch.backends.cudnn.deterministic = True
             torch.backends.cudnn.benchmark = False
 
-    def _setup_jax(self):
-        if self.config.deterministic:
-            jax.config.update("jax_default_matmul_precision", "highest")
-
-    def get_jax_rng(self):
-        return jax.random.PRNGKey(self.config.seed)
+    def _setup_python(self):
```
pglusacTT: This is a weird function name, not sure why we chose it, could you change it to something more meaningful?

pglusacTT (approved):
Left one comment otherwise good.

blacksmith/tools/reproducibility_manager.py
mmilosevicTT: Changed to `_seed_python_rngs`.

PR #708, opened 2026-08-12
Refactor import of `wandb` in logger

mmilosevicTT:
### Issue
N/A

### Problem description
WandB was imported at the top of the file even though it is only used when `use_wandb` is True. This import requires wandb to always be installed with trainer class even though we are not using it with value set to  true.

### What's Changed
Scope import of wandb only if the config is set.

### Checklist

### DIFF
--- a/blacksmith/tools/logging_manager.py
+++ b/blacksmith/tools/logging_manager.py
@@ -20,13 +20,16 @@ GOLDEN_LOGS_DIR = Path("tests/golden_files")
 def _import_wandb():
     """Import wandb only when W&B logging is enabled.
 
-    Keeps wandb an optional dependency for stdout-only consumers
-    (e.g. tt-media-server sets use_wandb=False).
+    Keeps wandb an optional dependency for stdout-only consumers.
     """
     try:
         import wandb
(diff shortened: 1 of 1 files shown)


blacksmith/tools/logging_manager.py
```diff
@@ -10,18 +10,31 @@
...
+def _import_wandb():
+    """Import wandb only when W&B logging is enabled.
+
+    Keeps wandb an optional dependency for stdout-only consumers
+    (e.g. tt-media-server sets use_wandb=False).
+    """
+    try:
+        import wandb
+    except ImportError as e:
+        raise ImportError("wandb is required when use_wandb=True. " "Install it with `pip install wandb`.") from e
```
pglusacTT: "Install it with `pip install wandb`" the user should probably use our requirements file if we need a specific version of wandb.

pglusacTT (approved):
One comment otherwise good!

blacksmith/tools/logging_manager.py
mmilosevicTT: Changed to suggestion of using our env.

PR #709, opened 2026-08-13
[Trainer] Fix Callback Calls in Trainer Class

mmilosevicTT:
### Issue
N/A

### Problem description
MetricsCallback was breaking graph in the point between forward and backward graph.

### What's Changed
This PR fixes:
- Double `zero_grad` on optimizer
- Move `loss.item()` accumulation from forward_end to backward_end.

### Checklist

### DIFF
--- a/blacksmith/tools/trainer/callbacks/metrics_callback.py
+++ b/blacksmith/tools/trainer/callbacks/metrics_callback.py
@@ -42,8 +42,6 @@ class MetricsCallback(Callback):
             trainer.logger.watch_model(trainer.model)
 
     def on_backward_end(self, trainer, loss, *args, **kwargs):
-        # After the trainer's post-backward sync so ``loss.item()`` does not force
-        # a forward-only XLA execute before backward is attached to the graph.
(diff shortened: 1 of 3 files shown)


blacksmith/tools/trainer/callbacks/metrics_callback.py
```diff
@@ -41,7 +41,7 @@ def on_train_start(self, trainer, *args, **kwargs):
         if trainer.config.logging.model_to_wandb:
             trainer.logger.watch_model(trainer.model)
 
-    def on_forward_end(self, trainer, loss, *args, **kwargs):
+    def on_backward_end(self, trainer, loss, *args, **kwargs):
```
pglusacTT, line 44: Why is this renamed to `on_backward_end`? What happened to `on_forward_end`? :D

mmilosevicTT: Both are present in base, we just moved where we do loss accumulation in MetricsCallback.

pglusacTT (approved):

PR #727, opened 2026-08-25
[Trainer] Perf Improvement for Trainer LoRA LLM Training

### LINKED ISSUE #724: Enable the existing, more performant Llama-3.1-8B training path via the trainer
A more performant training implementation for Llama-3.1-8B already exists in tt-blacksmith but is not wired into the workflow. Enable it so Llama-3.1-8B fine-tuning runs through this optimized path instead of the current one.

mmilosevicTT:
### Issue
Closes https://github.com/tenstorrent/tt-blacksmith/issues/724

### Problem description
Currently `Trainer` class is following a simple training structure without different workarounds that team made for perf improvements.

### What's Changed
Introduce following concepts to Trainer class:
- Fwd + Bwd + Opt as one graph
- Initialization of gradients and optimizer states
- Addition of `Trainer` experiment CI test
- Optimizer Level 1 for Trainer

### Checklist
- [x] CI passing (https://github.com/tenstorrent/tt-blacksmith/actions/runs/32861140824)


blacksmith/tools/cli.py
mcupacTT, line 21: Since `TrainerConfig` has nested `metrics.steps_freq`, `checkpoint.save_strategy`, `logging.use_wandb`, they won't be overtaken  from `_TEST_MODE_DEFAULTS` in the tests and the config defaults will be used instead of test defaults - therefore we need to handle this case

tests/training_test_cases.py
```diff
@@ -156,6 +156,20 @@
...
+            "timeout": 5000,
+        },
+        marks=[
+            pytest.mark.uplift,
+            pytest.mark.p150,
+            pytest.mark.torch,
+            pytest.mark.single_chip,
+        ],
+        id="tt-llama_3_1_8b-sst2-p150-trainer",
+    ),
```
mcupacTT, line 172: Just a comment for both of us: When we decide to merge this test to shield CI, we will need to add more thorough testing, also with resuming from later checkpoints

mcupacTT (approved):
Apart from one thing to fix, LGTM!

blacksmith/tools/cli.py
mmilosevicTT: Fixed 😄

tests/training_test_cases.py
mmilosevicTT: Agreed, but I would leave that as a part of shield integration PR, once we decide how many tests we want to actually add.

